PR #8, opened 2022-09-19
Added dropdown for hamburger menu

rayone121:


PickleNik:
it looks okay-ish as of 24 days ago

although sometimes stretches on navigation, cuz height is percentage based, not `vh`


but it breaks with current changes, again cuz height is percentage based so with current changes height is `0` (100% - 300px = <0px)


it could be "fixed" by not setting a height at all, but I would still play around with padding and overall styling

PickleNik (changes requested):
```diff
- .navbar-list.mobile{
+ .navbar-list.mobile {
  background-color: var(--grey-one);
  position: fixed;
  justify-content: center;
  display: grid;
- top:6rem;
+ top: 6rem;
- height: calc(100% - 300px);
  width: 25rem;
  border-radius: 2rem;
}
```

PR #12, opened 2022-10-12
feat: add contributors heading.

Aunali321:


src/routes/credits/+page.svelte
```diff
@@ -4,9 +4,47 @@
 </script>
 
 <div class="wrapper contrib-grid">
+    <h1>
+        <span class="redder">Re</span>Vanced Contributors
```
afnzmn, line 8: Use just 'Contributors' instead, since we say ReVanced everywhere else.

src/routes/credits/+page.svelte
```diff
@@ -4,9 +4,47 @@
...
+	}
+    h1 {
+        text-align: center;
+		color: var(--white);
+		font-weight: 700;
+		font-size: 5rem;
+		letter-spacing: -0.04em;
+		line-height: 1em;
+	}
+    .contrib-grid {
```
afnzmn, line 29: .contrib-grid isn't needed, I don't know why, but just remove.

PR #18, opened 2022-10-30
fix: production 404 page

Axelen123:


Axelen123:
Half the fix was setting the vercel preset to SvelteKit instead of Svelte.

PR #19, opened 2022-10-31
feat: generate doc pages from markup

Axelen123:
- [x] Move some of the CSS from `documentation.scss`
- [x] Ability to set path with env var
- [x] CI build script
- [X] Ability to add other file formats.
- [x] Asciidoc support. 
- [x] Maybe make a small adapter-static wrapper for vercel?
- [x] Fix Navbar (easy)


Axelen123:
We may have to change asset handling later but that is going to have to wait until the docs repo actually gets some proper docs and ends up needing the feature.

PR #41, opened 2022-12-04
feat: brand new Modal component

mirgb:
Brand new modal component. Remove api settings page, and make it a component instead. Remove unused imports, add <p> styling, make page reload when clearing cache, export `default_base_url` in the `settings.ts`.


src/lib/components/atoms/Modal.svelte
```diff
@@ -0,0 +1,64 @@
+<script>
+    import { click_outside } from "$lib/utils"
+    import { fade } from 'svelte/transition';
+    let modalOpen = false;
+    export let icon = '';
+</script>
+
+<div role="dialog" aria-modal="true" class="modal">
```
Axelen123, line 8: Can you move `aria-modal` and `role="dialog"` to the modal container?
Also, `class="modal"` is confusing because it tricks you into thinking it is the actual modal...

src/lib/components/atoms/ApiSettingsButton.svelte
```diff
@@ -0,0 +1,123 @@
...
+		url = settings.default_base_url;
+	}
+	import Modal from '$lib/components/atoms/Modal.svelte';
+</script>
+
+<Modal icon="/icons/settings.svg">
+	<div class="settings-menu">
+		<h3>Configure the backend</h3>
+		<div class="url-stuff">
+			<h4>Api URL:</h4>
```
Axelen123, line 20:
```suggestion
			<h4>API URL:</h4>
```
Just to make it consistent with flutter manager :)

Axelen123 (changes requested):
Thank you for doing this!
Your modal is so much better than the unstyled route that came before it.

src/lib/components/atoms/ApiSettingsButton.svelte
```diff
@@ -0,0 +1,123 @@
+<script lang="ts">
+	import * as settings from '../../../data/api/settings';
+	import { clear } from '../../../data/api/cache';
+	let url = settings.api_base_url();
+	function handler() {
```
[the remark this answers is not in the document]
Axelen123: > This is never going to work now because we forgot to reorder this after making `clear()` reload. Speaking of that, the `clear()` should probably also be renamed to `clear_and_reload()` or something.

Ugh I was referring to the `reset()` function. GitHub UI moment

src/app.css
```diff
@@ -33,6 +33,17 @@ body {
...
+		background-color: transparent;
+		border: none;
+		font-family: inherit;
+		font-size: inherit;
+		font-style: inherit;
+		font-weight: inherit;
+		line-height: inherit;
+		padding: 0;
+}
+
```
Axelen123, line 46: formatting moment. I will fix with a separate commit.

Axelen123 (approved):
Finished :)

PR #48, opened 2022-12-23
feat: implement scrollbar on small screens on the navigation

WadieBenabdouh:
Create slide scrollbar for the navigation (Home, download and patches...) for devices with small screen widths 300px - 414px.


Sculas (commented):
It works, but the position of the Settings icon is too far to the left. Also, the colors on the right border seem a little off, are they supposed to be there? 
image

xafn:
This would only be a temporary solution, currently working on a mobile navigation menu instead. But thank you for the contribution!

PR #70, opened 2023-02-17
fix: resolve all issues in `npm run check`

xafn:
- [x] need to fix documentation related errors  please ax

### DIFF
--- a/src/routes/docs/DocsNavTree.svelte
+++ b/src/routes/docs/DocsNavTree.svelte
@@ -1,12 +1,19 @@
 <script lang="ts">
-	import { is_tree } from './documentation.shared';
-	import type { DocsTree } from './documentation.shared';
+	import { is_tree, type DocsTreeNode } from './documentation.shared';
+	import type { DocsTree, DocumentInfo } from './documentation.shared';
 
 	import DocsNavNode from './DocsNavNode.svelte';
 
 	export let tree: DocsTree;
 	// How deeply nested this is.
 	export let nested = 0;
+
+  function assert_info_node(v: DocsTreeNode) {
+    if (!is_tree(v)) {
+      throw new Error("not tree aaa");
+    }
+    return v as DocumentInfo;
+  }
 </script>
 
 {#if nested}
@@ -25,7 +32,7 @@
 			<!-- Recursion here is fine. We are not dealing with a tree the size of a linux root file system. -->
 			<svelte:self tree={node} nested={nested + 1} />
 		{:else}
-			<DocsNavNode info={node} />
+			<DocsNavNode info={assert_info_node(node)} />
 		{/if}
 	{/each}
 </ul>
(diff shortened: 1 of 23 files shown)


xafn:
Also, would it be possible to move `/images` into `/static/images`? @Axelen123

Axelen123:
> Also, would it be possible to move `/images` into `/static/images`? @Axelen123

I think so. Just add an import alias

src/routes/download/+page.svelte
```diff
@@ -23,7 +23,7 @@
 	<p>Patch your favourite apps, right on your device.</p>
 	<div class="buttons">
 		<Query {query} let:data>
-			<Button kind="primary" icon="download" href={data.assets[0].url} target="_blank">
+			<Button kind="primary" icon="download" href={data.assets[0].url} download>
```
Axelen123, line 26: mimimi the button needs target=_blank or rel=external (will mess up the sveltekit router otherwise).
I don't think download does anything (or does it?).

Axelen123 (changes requested):

src/routes/download/+page.svelte
xafn: It actually does do something, instead of opening a new tab and closing it, download prop just downloads the file on the same tab.

Axelen123: oh ok then

Axelen123:
> * [ ]  need to fix documentation related errors  please ax

I will be able to help you when my fever debuff expires

xafn:
> > * [ ]   need to fix documentation related errors  please ax
> 
> I will be able to help you when my fever debuff expires

Hope you get better soon, and no rush either.

src/routes/docs/DocsNavTree.svelte
```diff
@@ -1,12 +1,19 @@
...
+	import type { DocsTree, DocumentInfo } from './documentation.shared';
 
 	import DocsNavNode from './DocsNavNode.svelte';
 
 	export let tree: DocsTree;
 	// How deeply nested this is.
 	export let nested = 0;
+
+  function assert_info_node(v: DocsTreeNode) {
+    if (!is_tree(v)) {
```
xafn, line 12: Wouldn't this function always be `!is_tree` because of the if-else statement?

xafn: @Axelen123 help the error you threw is whats not letting me merge gah whys it there

Axelen123: lol sorry I forgor.
I will check this in a few hours

Axelen123: I think it is best to move it to the docs file. I will do that real quick.

Axelen123: The reason why we need the function is because the Array has both types of nodes. We get the nodes from the `each` block in the template but it is not possible to use `as` outside script tags even though we know it is an info node. I added the check just in case someone tries to use it incorrectly.

Axelen123 (approved):

Issue #71, opened 2023-02-23
fix: prevent access without authorization token to `/poll`


oSumAtrIX:
## Issue

Currently, https://revanced.app/poll can be visited without providing any access token.

## Solution

Prevent access by either returning a custom page explaining, why access is denied and how to gain access, redirect to main page or return an error page with code 401.

Axelen123:
I do not see why this is a problem right now, since I doubt anyone will navigate there without a token from the bot.

oSumAtrIX:
The website currently does not have any relationship with Discord or the bot. Anyone can issue a token to acess that page. As an example, the token could be issued over Telegram or Google and so on. The website is only aware of that it needs a token. For that reason it should also handle the case when no token is provided, regardless of where this token has been issued (because it doesn't matter for the page).

Axelen123:
https://github.com/revanced/revanced-website/commit/575ea74ab504ed839c93fdc0a799b36a435eb5a7

PR #74, opened 2023-02-23
feat(poll): prevent submission, when there is no selection

### LINKED ISSUE #72: fix: prevent poll submission, when there is no selection
## Issue

Currently, you can press the "Submit" button even when there is no selection.

## Solution

Disable the "Submit" button, when there is no selection or round-robin, until at least a selection was made when the end was reached.

Ushie:
Fixes #72

### DIFF
--- a/src/lib/components/atoms/LogoOption.svelte
+++ b/src/lib/components/atoms/LogoOption.svelte
@@ -3,9 +3,9 @@
 	import Modal from './Dialogue.svelte';
 	import Button from './Button.svelte';
 
-  import next from '$lib/assets/icons/next.svg';
-  import previous from '$lib/assets/icons/previous.svg';
-  import type { APILogo } from '$lib/types';
+	import next from '$lib/assets/icons/next.svg';
+	import previous from '$lib/assets/icons/previous.svg';
+	import type { APILogo } from '$lib/types';
 
 	export let name = '';
 	export let selected: string[];
@@ -171,11 +171,6 @@
 		font-weight: 800;
 	}
 
-	h6 {
-		font-size: 0.9rem;
-		overflow: hidden;
-		text-overflow: ellipsis;
-	}
 	.text {
 		width: 100%;
 		white-space: nowrap;
@@ -186,8 +181,7 @@
 		background-color: var(--accent-low-opacity);
 	}
 
-	.clicked h2,
-	.clicked h6 {
+	.clicked h2 {
 		color: var(--white);
 	}
 
--- a/src/routes/poll/+page.svelte
+++ b/src/routes/poll/+page.svelte
@@ -2,15 +2,15 @@
 	import { onMount } from 'svelte';
 	import { fly } from 'svelte/transition';
 	import { expoOut } from 'svelte/easing';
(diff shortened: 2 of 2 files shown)


oSumAtrIX (approved):
While this PR solves #71, it does not help the user to solve this issue on their end. Basically, the user has to go back and make a selection, and the "flow" breaks. Minor issue but worth consideration regardless in favor of UX.

Ushie:
How should that be resolved? should a button be added to the popup to go back to Page 1?

Ushie:
Or did you mean in terms of text clarity/explanation

Ushie:
Got it, should the "review selection" button always be displayed? or should it only be displayed after they view all the pages for their first time

oSumAtrIX:
At least after the first selection, but making it show up after all logos have been reviewed sounds like a good idea.

Ushie:
Also I'm unsure so I'll ask, "round-robin" would be simply going back to the first page if there has been no selection made? Would you rename the "Next" button to "Go back" or keep it as "Next"? I feel like Next would be better but would that give an implication that the logos will be different?

oSumAtrIX:
Basically:
1. Queue all submissions into a FIFO list and mark the last pushed one as the "last" submission
2. Pop X amount and display them to the user
3. User makes their selection; push the rest of the unselected submissions back to the list
4. When the submission marked as "last" is poped but the user has not selected any submission repeat with `2.` at the moment the user makes a selection (or if they made one before "last" was poped), display the button to review the submission and allow the user to commit it.

Ushie:
Would that not be accomplished easier if we just:

```diff
	function nextPage() {
		if (currentPage >= logoPages || submit) return null;
		currentPage++;

		min = currentPage * logoAmount;
		max = min + logoAmount;

		if (currentPage < logoPages) {
			const nextPage = currentPage + 1;
			const nextMin = nextPage * logoAmount;
			const nextMax = nextMin + logoAmount;
			logos.slice(nextMin, nextMax).forEach(({ variants }) => {
				variants.forEach((variant) => preloadImage(variant.gdrive_direct_url));
			});
		}
+             if (currentPage = finalPage) {
+                 showSubmit = true;
+             }
		transitionDirection = 5;
	}```

Ushie:
Oh I see the point of your message, you want the logos that aren't selected not to show up on the round-robin, hm

oSumAtrIX:
In that case, the user will have to go back to reselect logos they might have been interested in before or when they have not made any selection.

Axelen123:
@Ushie could you make the button unclickable if nothing has been selected?

Ushie:
I'll continue working on this tomorrow after school if things go well 👍

Ushie:
I'll continue working on this tomorrow after school if things go well 👍

Ushie:
TODO:

- [ ] Next round-about
- [ ] Previous round-about after first next round-about
- [ ] Allow review submission after reaching the last page once
- [ ] Option to go back to viewing logos when in review mode

Axelen123 (changes requested):

Axelen123 (approved):

PR #81, opened 2023-05-13
fix(Spinner): properly center align

### LINKED ISSUE #57: fix: Loading Animation alignment
Current Loading Animation is not centered on mobile screen. It's a little bit on right side.
Screenshot_2023-01-02-20-34-44-592_com chrome canary

https://user-images.githubusercontent.com/82371061/210246049-cee88498-62e3-410c-8f56-593abe081233.mp4

niyasrad:
### Description

- The current loader, is held to the right, which is due to the positioning of the spinner component.
- The styles have been calculated, changed accordingly to ensure that it's perfectly centered.

### Screenshot

Here's a screenshot to show the new loading position doing it's work! I set up grids to make sure that they're in the blind center.

image

### DIFF
--- a/src/lib/components/Spinner.svelte
+++ b/src/lib/components/Spinner.svelte
@@ -11,15 +11,22 @@
 		}
 	}
 
-	.spinner:before {
-		content: '';
-		box-sizing: border-box;
-		width: 50px;
-		height: 50px;
-		border-radius: 50%;
+	.spinner {
 		position: fixed;
 		top: 50%;
 		left: 50%;
+		width: 50px;
+		height: 50px;
+		transform: translate(-50%, -50%);
+	}
+
+	.spinner:before {
+		content: '';
+		box-sizing: border-box;
+		position: absolute;
+		width: 100%;
+		height: 100%;
+		border-radius: 50%;
 		border: 4.5px solid transparent;
 		border-top-color: var(--accent-color);
 		animation: spinner 0.6s linear infinite;


src/lib/components/Spinner.svelte
```diff
@@ -18,8 +18,8 @@
 		height: 50px;
 		border-radius: 50%;
 		position: fixed;
-		top: 50%;
-		left: 50%;
+		top: calc(50% - 25px);
+		left: calc(50% - 25px);
```
xafn, line 22: While this works, the more common way of doing it would be adding a `transform: translate(-50%, -50%)` style to the class. If the size of the spinner ever changed, we would only need to edit 2 values instead of 4.

xafn (changes requested):

src/lib/components/Spinner.svelte
niyasrad: Understood! Will try and implement it!

niyasrad: image

I've certainly fixed it that way, but i'll need your review for this

image

I've added a container over the spinner, and used `.spinner` instead of `::before` so that it can access the full width of the container. Is this good of a fix? 

Please let me know, your thoughts.

niyasrad: I've rebased and pushed the code, please review!

src/lib/components/Spinner.svelte
```diff
@@ -2,7 +2,9 @@
 	import { fade } from 'svelte/transition';
 </script>
 
-<div class="spinner" transition:fade={{ duration: 250 }} />
+<div class="spinner-container">
+	<div class="spinner" transition:fade={{ duration: 250 }} />
+</div>
 
 <style>
 	@keyframes spinner {
```
xafn, line 10: Instead of adding a spinner-container, we could just add the translate style to the `spinner` animation. We can combine both the translate and rotate styles in one transform style. Try doing making a separate `from` keyframe and then adjusting the `to` keyframe.

xafn (changes requested):

src/lib/components/Spinner.svelte
xafn: Also, sorry for the late reply!

niyasrad: I've went back to your described approach, of making translate -50%, -50%, please review!

xafn (approved):

xafn:
Thank you!

PR #91, opened 2023-06-06
feat: move social icons to nav header

PickleNik:
before / after: 
Screenshot 2023-06-06 at 12 53 48 AM
Screenshot 2023-06-06 at 12 53 58 AM

Screenshot 2023-06-06 at 12 55 19 AM
Screenshot 2023-06-06 at 12 55 28 AM


src/layout/Navbar/SocialHost.svelte
oSumAtrIX, line 1: May be deleted

oSumAtrIX (approved):

src/layout/Navbar/SocialHost.svelte
PickleNik: oh i thought i did my b

oSumAtrIX:
In mobile view, unless someone opens the drawer, the icons are hidden, this should be changed back, at least in the mobile view.

PickleNik:
> In mobile view, unless someone opens the drawer, the icons are hidden, this should be changed back, at least in the mobile view.

like dis?
Screenshot 2023-06-06 at 7 00 27 PM

oSumAtrIX:
How about the original place? Would that be labeled as inconsistent because, on desktop view, it is present in the nav bar?

xafn:
I'd rather it be in the drawer than the main mobile nav

PickleNik:
> I'd rather it be in the drawer than the main mobile nav

it's in both

PickleNik:
> How about the original place? Would that be labeled as inconsistent because, on desktop view, it is present in the nav bar?



> How about the original place? Would that be labeled as inconsistent because, on desktop view, it is present in the nav bar?

I have an idea

PickleNik:
Screenshot 2023-06-06 at 7 16 03 PM
@oSumAtrIX @xafn how is this?

PickleNik:
Screenshot 2023-06-06 at 7 33 43 PM
or bigger

PR #113, opened 2023-06-07
feat: telemetry consent modal

404-Program-not-found:
Added a cookie consent popup on the first visit
screenshot

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -36,17 +36,16 @@
 	let askCookieConsent = false;
 
 	// store a cookie to remember the user's preference
-	function setPrivacySetting(doCookie: boolean) {
-		localStorage.setItem('doCookie', doCookie.toString());
+	function setPrivacySetting(allowTelemetry: boolean) {
+		localStorage.setItem('allowTelemetry', allowTelemetry.toString());
 		askCookieConsent = false;
 		if (typeof telemetryInit !== 'undefined') {
-			//@ts-ignore don't worry about this, it will work
 			telemetryInit();
 		}
 	}
 
 	onMount(() => {
-		if (!localStorage.getItem('doCookie')) {
+		if (!localStorage.getItem('allowTelemetry')) {
 			askCookieConsent = true;
 		}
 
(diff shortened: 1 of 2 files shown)


oSumAtrIX:
The button actions should switch places. Left = yes, right = no

404-Program-not-found:
updated

kazimmt:
I think you should align the buttons on the Right side
Like this:
IMG_20230607_093049

404-Program-not-found:
> I think you should align the buttons on the Right side
> Like this:
> IMG_20230607_093049
> 

The position is purposefully not aligned the same so users don't spam through accept

oSumAtrIX:
> > I think you should align the buttons on the Right side
> > Like this:
> > IMG_20230607_093049
> 
> The position is purposefully not aligned the same so users don't spam through accept

I don't think it's right to motivate the alignment like that. Should be checked with M3

404-Program-not-found:
fair enough, will update in a sec

src/routes/+layout.svelte
```diff
@@ -71,25 +90,61 @@
...
 </svelte:head>
 
 <QueryClientProvider client={queryClient}>
 	<NavHost />
+	<!-- if cookie consent hasn't been set -->
+	<Dialogue bind:modalOpen={askCookieConsent} dismissible={false}>
+		<svelte:fragment slot="title">It's your choice</svelte:fragment>
+		<svelte:fragment slot="description">
+			We use analytics and cookies to help us better understand how you use the site. Opting in is
+			optional and won't impact your experience.
```
oSumAtrIX, line 141:
```suggestion
			This site uses analytics to understand better how you use it. Opting in is
			optional and won't impact your experience.
```

alexandreteles:
What is the reason for it to overlay the whole page? Cookie consent is purposefully put in the footer and without blocking navigation so it doesn't break neither UI or UX. It should also not block interaction with other elements so it doesn't break if adblockers hide it.

alexandreteles:
@xafn Did you remember to add the privacy and cookie pages before merging?

PR #120, opened 2023-06-11
feat: remove wavy text decoration

oSumAtrIX:
## About

This PR reverts the design choice of using wavy text decoration. 

## Issue

The reason for that is that it looks out of place and doesn't harmonize with the rest of the website's design. 

## TODO

- [ ] Highlight additionally using `background-color` instead of only `color`
- [ ] Consider wavy footer separator

### DIFF
--- a/src/layout/Footer/FooterHost.svelte
+++ b/src/layout/Footer/FooterHost.svelte
@@ -155,9 +155,6 @@
 	}
 
 	a:hover {
-		text-decoration: underline;
-		text-decoration-style: wavy;
-		text-decoration-color: var(--accent-color-two);
 		color: var(--white);
 	}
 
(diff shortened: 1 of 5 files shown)


PalmDevs:
In my opinion, I think the underline would be pretty nice. Normal underline though, not a wavy one.

oSumAtrIX:
@PalmDevs 

Without underline:

https://github.com/revanced/revanced-website/assets/13122796/0ec60617-f038-434a-b374-8ba7cc25705d

With underline:

https://github.com/revanced/revanced-website/assets/13122796/33b0f190-880e-4db8-99ec-2cd9f5bca205

Continuing here: https://github.com/revanced/revanced-website/pull/121

PR #121, opened 2023-06-11
feat: remove wavy text decoration

oSumAtrIX:
## About

Accidentally deleted the branch, continuing https://github.com/revanced/revanced-website/pull/120

This PR reverts the design choice of using wavy text decoration. 

## Issue

The reason is that it looks out of place and doesn't harmonize with the rest of the website's design. 

## TODO

- [ ] Highlight additionally using `background-color` instead of only `color`
- [x] Consider wavy footer separator

### DIFF
--- a/src/layout/Footer/FooterHost.svelte
+++ b/src/layout/Footer/FooterHost.svelte
@@ -155,9 +155,7 @@
 	}
 
 	a:hover {
-		text-decoration: underline;
-		text-decoration-style: wavy;
-		text-decoration-color: var(--accent-color-two);
+		text-decoration: underline var(--accent-color-two);
 		color: var(--white);
 	}
 
--- a/src/routes/contributors/+page.svelte
+++ b/src/routes/contributors/+page.svelte
@@ -83,9 +83,7 @@
 	}
 
 	a:hover {
-		text-decoration: underline;
-		text-decoration-style: wavy;
-		text-decoration-color: var(--grey-four);
+		text-decoration: underline var(--grey-four);
 	}
 
 	a:hover::after {
--- a/src/routes/contributors/ContributorPerson.svelte
+++ b/src/routes/contributors/ContributorPerson.svelte
@@ -26,9 +26,7 @@
 	}
 
 	a:hover {
-		text-decoration: underline;
-		text-decoration-style: wavy;
-		text-decoration-color: var(--accent-color);
+		text-decoration: underline var(--accent-color);
 		color: var(--white);
 	}
 
--- a/src/routes/contributors/ContributorSection.svelte
+++ b/src/routes/contributors/ContributorSection.svelte
@@ -79,9 +79,7 @@
 	}
 
 	a:hover {
-		text-decoration: underline;
-		text-decoration-style: wavy;
-		text-decoration-color: var(--accent-color);
+		text-decoration: underline var(--accent-color);
 		color: var(--white);
 	}
 
--- a/src/routes/patches/PatchItem.svelte
+++ b/src/routes/patches/PatchItem.svelte
@@ -92,10 +92,8 @@
 	}
 
 	a .patch-info:hover {
-		text-decoration: underline;
+		text-decoration: underline var(--accent-color-two);
 		color: var(--accent-color-two);
-		text-decoration-style: wavy;
-		text-decoration-color: var(--accent-color-two);
 	}
 
 	.info-container {


PalmDevs:
Continuing from the accidentally closed PR:

I think having an underline is generally better for accessibility. Colorblind people may not be noticing the color change if it was only the color that changed. Normal underlines should do the job.

src/layout/Footer/FooterHost.svelte
```diff
@@ -155,9 +155,6 @@
 	}
 
 	a:hover {
-		text-decoration: underline;
-		text-decoration-style: wavy;
-		text-decoration-color: var(--accent-color-two);
 		color: var(--white);
```
oSumAtrIX, line 158:
```suggestion
		text-decoration: underline;
		text-decoration-color: var(--accent-color-two);
		color: var(--white);
```

src/routes/contributors/+page.svelte
```diff
@@ -83,9 +83,7 @@
 	}
 
 	a:hover {
-		text-decoration: underline;
-		text-decoration-style: wavy;
-		text-decoration-color: var(--grey-four);
+		color: var(--white);
```
oSumAtrIX, line 86:
```suggestion
		text-decoration: underline;
		text-decoration-color: var(--grey-four);
		color: var(--white);
```

src/routes/contributors/ContributorSection.svelte
```diff
@@ -79,9 +79,6 @@
 	}
 
 	a:hover {
-		text-decoration: underline;
-		text-decoration-style: wavy;
-		text-decoration-color: var(--accent-color);
 		color: var(--white);
```
oSumAtrIX, line 82:
```suggestion
		text-decoration: underline;
		text-decoration-color: var(--accent-color);
		color: var(--white);
```

src/routes/contributors/ContributorPerson.svelte
```diff
@@ -25,13 +25,6 @@
...
 	}
 
-	a:hover {
-		text-decoration: underline;
-		text-decoration-style: wavy;
-		text-decoration-color: var(--accent-color);
-		color: var(--white);
-	}
-
 	h5 {
```
oSumAtrIX, line 28:
```suggestion
	a:hover {
		text-decoration: underline;
		text-decoration-color: var(--accent-color);
		color: var(--white);
	}

	h5 {
```

src/routes/patches/PatchItem.svelte
```diff
@@ -92,10 +92,7 @@
 	}
 
 	a .patch-info:hover {
-		text-decoration: underline;
 		color: var(--accent-color-two);
```
oSumAtrIX, line 95:
```suggestion
		text-decoration: underline;
		text-decoration-color: var(--accent-color-two);
		color: var(--accent-color-two);
```

src/routes/contributors/+page.svelte
```diff
@@ -84,8 +84,8 @@
 
 	a:hover {
 		text-decoration: underline;
-		text-decoration-style: wavy;
 		text-decoration-color: var(--grey-four);
+		color: var(--white);
```
oSumAtrIX, line 88:
```suggestion
```

PalmDevs (commented):
Not really a required change, but since you're already using the `text-decoration` property. Why not just put the color there in the same property because `text-decoration` is a shorthand for multiple properties.

```css
/* For example */
text-decoration: underline var(--accent-color);
```

PalmDevs (approved):
LGTM 😁

oSumAtrIX:
Check the TODOs from the description before that

PalmDevs:
> Check the TODOs from the description before that

TODO #1: Why do you need to highlight the background color as well. Underline and color change should be enough, like how Google does hyperlinks.

TODO #2: This can be another issue/PR.

oSumAtrIX:
> TODO 1: Why do you need to highlight the background color as well. Underline and color change should be enough, like how Google does hyperlinks.

It looks good and gives a stronger visual feedback which improves UX.

> TODO 2: This can be another issue/PR.

Alright

PalmDevs:
> It looks good and gives a stronger visual feedback which improves UX.
I feel like just an underline where the user hovers is just enough to tell which link you're hovering on. Maybe for keyboard navigation, a background color/outline may fit (style via `:focus` pseudo selector). But if the user is using a mouse, they'd already know which link they're hovering on. A color shift and an underline is already enough visual feedback. Entirely my opinion though.

PR #135, opened 2023-07-03
fix: use correct name in download button

oSumAtrIX:


xafn (approved):

oSumAtrIX:
It may just be me, but it looks like the label isn't properly centered. The reason for this is that the focus is on the label text. The text is shifted a little to the right to center the icon so it looks off. Artificial shifting to center the text could be used to work against this issue

PR #151, opened 2023-07-30
feat(schema.org): JsonLd

Ushie:


src/routes/+page.svelte
alexandreteles, line 1: I know JSON-LD is supposedly easier to implement, but microdata fits better in the HTML the way God intended it to be: markup first. Nevertheless, everything seems fine.

src/routes/contributors/+page.svelte
alexandreteles, line 1: You should probably implement an ItemList with Items of the Contributor type, but this is enough for now.

src/routes/download/+page.svelte
alexandreteles, line 1: Here the primary type should also be WebPage, and then the element that contains the download itself be marked as mainEntity with the link to the download being the significantLink. For now, this is fine but think about opening an issue requiring the download link to be accessible from the outer scope.

src/routes/patches/+page.svelte
alexandreteles, line 1: `Patch` is problematic; we need to review the available types... SoftwareSourceCode? In any case, those should also be enveloped in an ItemList like discussed on the Contributors page, but this can also wait; this is good for now.

alexandreteles (commented):
👍🏾

alexandreteles (approved):

Issue #147, opened 2023-07-22
fix: show all compatible versions


oSumAtrIX:
# Issue

Currently, the only version shown for patches on the `/patches` page is the "suggested" for each patch. This hides the important information of the patches other compatibilities to versions

# Solution

How all versions that patches are compatible with.

# Additional context

- Since the list can be quite big, collapse the list and show the suggested only. 
- When uncollapsing, uncollapse all the patches list of compatible versions, not just one, because the user's intention is likely to see them all. When collapsing again, hide them all again and show the suggested only.
- It might be favourable to index versions for search as well as add list filters for them

Ushie:
Do you imagine it to look like this? 
![image](https://github.com/ReVanced/revanced-website/assets/79272171/5df8e1db-3665-49eb-8e06-4639557d816f)
or what do you exactly envision

oSumAtrIX:
Every version has a chip. The intepretation of these chips are "tags".

PalmDevs:
> Do you imagine it to look like this? 
> ![image](https://github.com/ReVanced/revanced-website/assets/79272171/5df8e1db-3665-49eb-8e06-4639557d816f)
> or what do you exactly envision

This is definitely better than separate chips. For one, it includes all the information with the same hierarchy in the exact same place, and for two, it doesn't waste too much space like multiple chips.

oSumAtrIX:
> For one, it includes all the information with the same hierarchy in the exact same place, and for two, it doesn't waste too much space like multiple chips.

I disagree.

>  For one, it includes all the information with the same hierarchy in the exact same place

So do chips with a version in each chip

> , it doesn't waste too much space like multiple chips.

It wastes as much space as separate chips minus the padding. This is negligible.

In addition to that, readability drastically decreases: 
![image](https://github.com/ReVanced/revanced-website/assets/13122796/d1de82ed-9c03-4c1b-ae35-ac184cdfd82d)

In addition to that, functionality drastically decreases:
With separate chips, you can click on a chip, which will filter the patches list for that version. This way an interaction is added to the chip purposefully.

Ushie:
> With separate chips, you can click on a chip, which will filter the patches list for that version. This way an interaction is added to the chip purposefully.

Would that not require changing the behavior of the package chip to do the same thing? because it's odd to use version chips for filter and then use the package name chip to link to the application's page

PalmDevs:
[...]

PR #152, opened 2023-07-30
feat: show all compatible versions

Ushie:


Ushie:
@PalmDevs @oSumAtrIX Please confirm that the both of you are satisfied with this implementation of #147

PalmDevs:
Do you have any preview of this?

Ushie:
Yes, check the deployment that the ReVanced bot sent

src/routes/patches/PatchItem.svelte
```diff
@@ -36,20 +37,48 @@
...
+			{#if patch.compatiblePackages[0].versions.length > 1}
+				<li class="patch-info button" on:click={() => (showAllVersions = !showAllVersions)}>
+					<img
+						class="expand-arrow"
+						id="expand-versions"
+						style:transform={showAllVersions ? 'rotate(90deg)' : 'rotate(-90deg)'}
+						src="/icons/expand_more.svg"
+						alt="dropdown"
+					/>
+				</li>
```
oSumAtrIX, line 79: I suggest to move  out of <li>. The reason is that the semantics of the <li> here is that it is a chip, but the expand button is not a chip:

Before:
image

After:
image

PR #161, opened 2023-08-09
chore: update manager screenshot

kazimmt:


xafn:
Can you provide a screenshot with more space at the top where the notification bar would, like in the old screenshot?

PR #192, opened 2023-10-23
[PoC]: Sync documentation from ReVanced Docs to Docusaurus 

validcube:
This code is created for old Docusaurus repo, rewriting in progress

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,7 +15,9 @@ jobs:
     steps:
     - name: Checkout the repository
       uses: actions/checkout@v4
-
+      # with:
+      #   submodules: 'recursive'
+        
     - name: Pull documentation
       run: | # cp -rfL docs/* ../docs fail? // cp -rf repositories/*/docs ../docs/* work but worse
         git clone https://github.com/ReVanced/revanced-documentation rv-docs && cd rv-docs
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,42 @@
+name: Receive Dispatch and Pull Docs
+
+on:
+  workflow_dispatch:
+  repository_dispatch:
+    types: [docs]
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  pull-docs:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,42 @@
...
+  pull-requests: write
+
+jobs:
+  pull-docs:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout the repository
+      uses: actions/checkout@v4
+
+    - name: Pull docs from rv-docs repository
```
oSumAtrIX, line 19: Don't abbreviate here

```suggestion
    - name: Pull documentation
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,42 @@
...
+      run: | # cp -rfL docs/* ../docs fail? // cp -rf repositories/*/docs ../docs/* work but worse
+        git clone https://github.com/ReVanced/revanced-documentation rv-docs && cd rv-docs
+        git submodule update --init --recursive
```
oSumAtrIX, line 22: Can't you use an action? Would be more elegant.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,42 @@
...
+
+    - name: Push changes to origin
+      run: |
+        git config --global user.email "action@github.com"
+        git config --global user.name "GitHub Action"
+        git add .
+        git commit -m "Updated docs"
+        git push origin master
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
oSumAtrIX, line 41: A commit action can be used.

.github/workflows/docs.yml
validcube: I prefer to write my first iteration of my CI with pure commands

action can be made at later iterations.

Ushie:
The docusaurous branch is stale, as docusaurous may be separated, and a sync method hasn't been decided (there are some unsettled issues)

oSumAtrIX:
I'd say we can allocate an issue for this?

Ushie:
Sure, go ahead and open one as you were the one who raised the issue with this type of syncing

PR #209, opened 2023-12-29
refactor: Minify assets

KTibow:
minifies SVGs using a combination of SVGO and manual work
minifies PNGs using optipng
this should make the website load faster as less data is being sent to the client
fingers crossed nothing broke


oSumAtrIX:
Please explain what this PR aims to do and what "mini dying svgs" means

KTibow:
> what this PR aims to do

Make images using the SVG format smaller by removing unnecessary material.

> and what "mini dying svgs" means

In the context of websites, "minify" refers to the process of making files smaller by removing unnecessary data without affecting their functionality or display. When applied to SVGs (Scalable Vector Graphics), minification involves removing unnecessary metadata, comments, empty tags, and other non-essential information that might have been included during the creation or editing process.

Minifying an SVG can improve the performance of a website by reducing the amount of data the client's browser has to download and process. This is particularly valuable for improving load times and reducing bandwidth usage. The visual quality of the SVG remains the same even after minification, as the process only strips out information that doesn’t affect the rendering of the image.

Ushie (changes requested):
This breaks the wavey line in FooterHost

Ushie (approved):

PR #216, opened 2024-01-08
feat: Add verified badge to team member card

oSumAtrIX:
# About

This PR builds on https://github.com/ReVanced/revanced-api/issues/150 to display GPG keys for members:

image

# Todo

- [x] Merge https://github.com/ReVanced/revanced-api/issues/150 to main
- [x] Remove hardcoded keys for testing
- [x] Fix DOM semantics
- [x] Properly centre tooltip
- [x] Use correct M3 guidelines for tooltip
- [x] Add a way for mobile users to view the tooltip
- [x] Fix overflow on narrow screen space


PalmDevs:
Design wise, it looks pretty okay. Although, I think it'd be better to put the keys into a `<ul>` or `<ol>` and align all text to the left.

Usability isn't that good right now but you've already noticed the issue, so I won't repeat the same thing.

oSumAtrIX:
GPG keys can be very long, this has to be dealt with as well

KTibow (commented):
This seems like a bad place to put this info.

If I was looking for info about the team, I would probably open "contributors". not the donate page. If I was looking to donate, I wouldn't care about your GPG keys.

I would say more (eg suggest linking to `https://github.com/user.gpg`) but I won't because I don't understand the underlying motivation for
a. putting the GPG keys there in the first place (and to that extent, showing that a user is verified)
b. showing who the team is on the donate page

oSumAtrIX:
Moving the badge to the contributors tab leaves no proper place for the badge to be shown and it gets lost between all the other contributors or you'd have to display it for all of them which isn't the the purpose of adding the badge.

The donation page was chosen as it is currently the only place the website hits the members endpoint which returns information about the team. It is done on the donation page to see who you are donating too. Look up GitHub Sponsors for a native implementation of that. 

To see the motivation to why GPG keys are added to users, check on this: https://github.com/ReVanced/revanced-api/issues/150#issuecomment-1881897951

oSumAtrIX:
I am not sure what the exact difference is between the key visible in your suggested URL and the one the API will return. Users can have a list of keys which is why I don't know how that would work on your URL. Apart from that, it seems like your URL returns the full public key whereas the API will return some short form of it. If there is a reason to use your URL over what is returned by the API. Please give it a comment with a reason on the issue I sent in my previous comment. In the case we return that, we either may want to separate the endpoint as it is gonna take up most of the requests size but usually not be necessary. Additionally, the website would have to rethink how to display it.

oSumAtrIX:
a. [As referenced](https://github.com/ReVanced/revanced-api/issues/150#issuecomment-1881828902) in one my previous comment, it is a common security practice. If you trust the domain revanced.app, you can transitively trust keys published there. If someone signs something with one of these keys, you can transitively trust them. An example scenario would be that you can prove in an arbitrary place on the web that you are who you claim to be on revanced.app. Inheritently, the trust chain starts at revanced.app and ends with your sign, which means whoever revanced.app claims this key belongs to is not verifiable without any additional means. Because it is not important who is behind the key, and instead it is only important that you are someone trusted on revanced.app, the keys are published as a verification method.

KTibow:
Okay, fair enough. Maybe in some later day contributors/team/donate would have a different organization but I'm fine with putting them there.

Now on the topic of displaying them, I think it could work better to have a dialog triggered either by a separate button or clicking on the badge instead. What's your opinion on that?

oSumAtrIX:
> Okay, fair enough. Maybe in some later day contributors/team/donate would have a different organization but I'm fine with putting them there.

If you want please open an issue. Open to suggestions :)

> Now on the topic of displaying them, I think it could work better to have a dialog triggered either by a separate button or clicking on the badge instead. What's your opinion on that?

I think it is expected to hover on the badge to open the tooltip. If a popup were to be shown, a click should be used. On mobile, the badge can be expanded to look like this (with the icon for example):

image

KTibow:
Okay, sounds like a plan. I have to go but when I get back I guess I can take a look at the code.
(One more thing, given I'm assigned are you expecting me to a. leave a review as is b. wait until it's finished and leave a review then c. update the code to finish it?)

oSumAtrIX:
It would be good if you could look into resolving the todos and push it. I have had my attempts but didn't really succeed in them

Ushie:
Need ideas for mobile support

oSumAtrIX:
On mobile, next to the badge show a label

Ushie:
@oSumAtrIX Please fix our Cloudflare deployment credentials

Ushie:
Heads up, this depends on https://github.com/ReVanced/revanced-api/pull/188 being merged

Issue #224, opened 2024-01-24
Changes required in the "Search for patches" placeholder in patches tab


RiddhimannM:
Listing down few issues/enhancement ideas for the search for patches placeholder in the patches path based on my observations as an end-user.

PS - This is my first issue as an open source contributor, so please excuse me for mistakes if any :)

- [x] 1. Currently not able to search a keyword containing hyphen {'-'} in it. See video for reference.

https://github.com/ReVanced/revanced-website/assets/84954173/b7552c85-622f-4da4-8715-fac7d1cd9af6

- [ ] 2. The search bar getting disappeared from view while I scroll down the list of patches. Would prefer a better UX layout which will move the searchbar to a location accessible even while scrolling the page.

https://github.com/ReVanced/revanced-website/assets/84954173/adc50a10-9e9d-47e3-9cd3-a5b2fae6ebcf

oSumAtrIX:
1. It looks like a string sanitization problem. In older times the patch names were stripped of the "-" character, so that may be the cause.
2. This would need some semantical changes such as moving the search from the header to the content

oSumAtrIX:
1. is fixed, please reopen an issue for 2.

PR #225, opened 2024-01-24
fix: handle hyphens properly in patches

KTibow:
- make query sanitization a common function
- also remove hyphens from each patch name (see #224)

### DIFF
--- a/src/routes/patches/+page.svelte
+++ b/src/routes/patches/+page.svelte
@@ -22,6 +22,12 @@
 	import Query from '$lib/components/Query.svelte';
 
 	const query = createQuery(['patches'], queries.patches);
+	const sanitize = (str: string) =>
+	  str.replace(/\./g, '')
+			.replace(/\s/g, '')
+			.replace(/-/g, '')
+			.replace(/_/g, '')
+			.toLowerCase();
 
 	let searchParams: Readable<URLSearchParams>;
 	if (building) {
@@ -32,12 +38,7 @@
 
 	$: selectedPkg = $searchParams.get('pkg');
 	let searchTerm = $searchParams.get('s');
-	let searchTermFiltered = searchTerm
-		?.replace(/\./g, '')
-		.replace(/\s/g, '')
-		.replace(/-/g, '')
-		.replace(/_/g, '')
-		.toLowerCase();
+	let searchTermFiltered = searchTerm && sanitize(searchTerm);
 
 	let timeout: ReturnType<typeof setTimeout>;
 	let mobilePackages = false;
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
Is the sanitization necessary? A better fit for searching would be fuzzy search but out of scope for this PR

KTibow:
the sanitization was already in place. i just moved it around a bit.

oSumAtrIX:
The function should be removed. It was used when patches were named-like-this

KTibow:
i don't think it should. both "wifi" and "wi-fi" should match "wi-fi", and the only way to do that is by sanitizing both.

KTibow:
I see your point that `-` is arbitrary, but I don't follow the argument that we must switch to fuzzy search. How do you feel about switching to `str.replace(/\W/g, '')` for both the query and patch details?

oSumAtrIX:
Fuzzy search is particularly useful if you want to match strings with any kind of mistake stochastically. This fixes your attempt at sanitizing the string so that wifi matches Wi-Fi for example but expands the character set to any kind of character. Any other sanitization attempt would be inferior to that including your replacement script as you only sanitize a subset now. Like I said, the suggestion to use fuzzy search is a follow up on you wanting to sanitize in the first place, but approached a naive solution, whereas this PR is not even about sanitization or fuzzy search but instead supposed to be a fix for the referenced issue as to why, if you still plan on this idea, a separate issue for discussion and PR should be made

KTibow:
While I understand the use for fuzzy search, I don't understand why you think that `\W` is a "subset" - it quite literally removes *everything* except for letters, numbers, and underscores. Yes, this PR's way of going about it was wrong, but I'm asking you because if you approve of using `\W` I'm going to change it to do so.

oSumAtrIX:
The reason for this is the same as before that this kind of approach is inferior to a fuzzy search. Fuzzy search already incorporates matching strings even when whitespace is left out or included just like your regex. What your regex can't, unlike fuzzy search, and therefore is inferior, is to match a string with a typo or anything other than your whitespace. Therefore your approach is still a subset of what Fuzzy search is. If we formulate the problem query "How to solve finding the patch even with mistakes such as including whitespace accidentally or making a typo" the best and complete solution to that formulation is not your regex, but fuzzy search. If implementation is a question, it should be as simple as using an npm package that would provide you with a fuzzy comparator.

KTibow:
Okay then, if you're sure that the only way this problem should be solved is with fuzzy search then you can go ahead and make an issue and PR for that as you described.

oSumAtrIX:
The issue referenced in this PR is primarily solved by removing the sanitization function that was left accidentally in the code. As explained earlier fuzzy search was a follow up on your intention to introduce sanitization as a concept by recycling the function that wasn't supposed to exist to begin with, and not something I initiated. I don't plan on making a PR for this therefore.

Issue #226, opened 2024-02-13
feat: Event-based theme


validcube:
This is just a little quirky idea but . . .

Hey, imagine this; tomorrow is 14th February (as of writing) and suddenly the colour accent of our website suddenly turned whoosh into a pink colour accent. Imagine in 1st January the website use a yellowish colour accent.

Imagine on April fool the website changed logo to the old one.

oSumAtrIX:
Not a bad idea. How easy would this be? I think the root colors can be modified programmatically, but some images have the colors hardcoded.

madkarmaa:
> I think the root colors can be modified programmatically

```js
document.documentElement.style.setProperty('--some-css-variable', '#696969')
```

PR #227, opened 2024-02-22
feat: Apply themes on specific events

madkarmaa:
## What is this?

Read #226 

## Description

I've created a simple JS library called [**DateTrigger**](https://github.com/madkarmaa/DateTrigger), which executes given code on given dates, just like the issue describes. The library is now added to the dependencies, and used in `+layout.svelte` so the changes can be globalized. While the library itself doesn't do anything on its own, it got supplied with a default set of events which, once again, fullfill #226 at least partially (changing CSS color variables). The variables selection can be tweaked according to your liking, and if the default events aren't enough or don't feel right, [custom events can be created](https://github.com/madkarmaa/DateTrigger?tab=readme-ov-file#basic-usage).

## How to test if it works?

Change the parameter passed to `DateTriggerEventsHandler` with the following:

```js
{
	TEST_EVENT: new DateTriggerEvent(new Date(), 10, () => {
		document.documentElement.style.setProperty('--text', '#e6e8fe');
		document.documentElement.style.setProperty('--accent', '#f62e65');
		document.documentElement.style.setProperty('--secondary', '#960670');
		document.documentElement.style.setProperty('--background', '#010215');
		document.documentElement.style.setProperty('--primary', '#7082f9');
	})
}
```

The event should fire immediately.

### ~~A little concern~~ fixed in f3aeadc

~~The library cannot work (**with default events**) if it doesn't have access to the `document` object, but using `browser` from `$app/environment` is really slow, at least on my end.~~


package.json
```diff
@@ -41,6 +41,7 @@
 		"@tanstack/query-core": "^4.36.1",
 		"@tanstack/query-persist-client-core": "^4.36.1",
 		"@tanstack/query-sync-storage-persister": "^4.36.1",
-		"@tanstack/svelte-query": "^4.36.1"
+		"@tanstack/svelte-query": "^4.36.1",
+		"datetrigger": "github:madkarmaa/DateTrigger"
```
KTibow, line 45: please use npm so we have stability, audibility, etc

src/routes/+layout.svelte
```diff
@@ -23,6 +23,11 @@
 	import { staleTime } from '$data/api';
 	import RouterEvents from '$data/RouterEvents';
 
+	// @ts-ignore
+	import { DateTriggerEventsHandler, DateTriggerEvent } from 'datetrigger';
+	// @ts-ignore
+	import { DEFAULT_EVENTS } from 'datetrigger/presets';
```
KTibow, line 29: why the ts-ignores? does your library not include typedefs?

src/routes/+layout.svelte
```diff
@@ -42,6 +47,9 @@
 	}
 
 	onMount(() => {
+		// must be running in the browser to interact with the document
+		new DateTriggerEventsHandler(DEFAULT_EVENTS);
```
KTibow, line 51: won't this cause style flashes while loading?

src/app.scss
```diff
@@ -38,15 +38,15 @@ body {
...
+	--bg-color: var(--background, hsl(252, 10%, 11%));
 	--grey-one: hsl(210, 14%, 17%);
 	--grey-two: hsl(212, 19%, 19%);
 	--grey-three: hsl(221, 17%, 26%);
 	--grey-four: hsl(226, 48%, 18%);
-	--grey-five: hsl(208, 30%, 75%);
+	--grey-five: var(--text, hsl(208, 30%, 75%));
 	--grey-six: hsl(230, 9%, 13%);
 	--grey-seven: hsl(240, 9%, 13.5%);
 	--grey-eight: hsla(207, 30%, 75%, 0.577);
```
KTibow, line 52: i don't like how this only replaces some colors, it would be better if there was a supplementary hue variable so that things like `--white` don't feel off color

package.json
madkarmaa: I'll have to publish a version, I'll do that

src/routes/+layout.svelte
madkarmaa: nope, plain js, the repo is linked in the pr description

src/routes/+layout.svelte
madkarmaa: yea, there's no other way to do it, at least not of my knowledge

src/app.scss
madkarmaa: custom events can be created

src/routes/+layout.svelte
KTibow: put a <script> tag in the html - it's standard practice if you must have custom theming

src/routes/+layout.svelte
KTibow: adding ts annotations would be appreciated

madkarmaa: then I'm gonna draft this pr and learn typescript

src/routes/+layout.svelte
madkarmaa: osumatrix told me to try and use npm instead of a script tag, but sure, if that's the only way, I'm down

madkarmaa:
# TODO

- [ ] Convert to TypeScript
- [ ] Make sure compiling TS outputs types
- [ ] Move code to a `<script>` tag in `app.html`

src/routes/+layout.svelte
madkarmaa: you mentioned to move the code in a script tag, is typescript needed in this case?

package.json
madkarmaa: done in 890fc7e

src/routes/+layout.svelte
madkarmaa: image shows the request to the url is successful, but the file is empty, even tho I can reach it and it clearly has contents
image

madkarmaa: using unpkg works, idk why gh doesn't. is this what you mean?
image

oSumAtrIX: I have explained why it is a better idea to use NPM.

It is not a good idea to load an external script through a <script>. For one it makes a outgoing request, for two, you don't pin the version and although possible, it would be better to bundle it with the current build for speed.

madkarmaa: ok but is there any way to have the same speed as a script tag? because even tho onMount is fast, it's not _as_ fast as a script, which is almost instantaneous

oSumAtrIX: I am not seeing how loading an external file, unoptimized for the current project at compile time and more, is faster. The remaining reasons also argue against script.

madkarmaa: I have no clue why,bill test again later on a different network, but the speed difference is very noticeable, the CSS changes are almost instantaneouswith the script tag, but with onMount they're delayed almost a whole second

oSumAtrIX: The technicality with onMount is not at question here. Its about network speed.

src/routes/+layout.svelte
KTibow: well in that case, you end up having to manually bundle the code and insert it as a string, so correct

src/routes/+layout.svelte
KTibow: @osumatrix I think you misunderstood a bit.

I'm not saying to load any libraries from the script tag. I'm saying that the script should be inlined in the HTML so there's no network delay. As stated before this is *the* way to prevent theme flashes.

madkarmaa: @KTibow 
> I'm saying that the script should be inlined in the HTML so there's no network delay. As stated before this is _the_ way to prevent theme flashes.

if I put the script tag in the head and leave the imports in +layout.svelte then it won't work

image
image

KTibow: Yup, imports aren't shared globally. You would need to inline it here.

madkarmaa: @KTibow doesn't work anyways :P
image
image

KTibow: As I stated, the imports aren't shared. The import won't magically work its way into the script tag, as you might be used to. Each script tag is treated as its own module thanks to bundling.

madkarmaa: @KTibow I know but what I don't know is how to "inline" it like you told me to

madkarmaa: @KTibow I know but what I don't know is how to "inline" it like you told me to

KTibow: Re: bundle size

While this library is great and has a logical structure, it would probably need to be simplified to reduce its size.
4.8kb minified / 2.2kb gzip'd is a bit large for something that could be represented as
```js
(() => {
const colors = {
  new_year: {text: "#e6e8fe", background: "#010215", primary: "#7082f9", secondary: "#960670", accent: "#f62e65"},
  valentines_day: {text: "#efd6f9", background: "#18051f", primary: "#dc8fef", secondary: "#93166c", accent: "#e34190"},
  april_fools: {text: "#ceeafe", background: "#000910", primary: "#7ec2fc", secondary: "#9504af", accent: "#fa28a7"},
  halloween: {text: "#f2eeed", background: "#0c0705", primary: "#df9d87", secondary: "#923112", accent: "#fd490e"},
  christmas: {text: "#fdfcfc", background: "#030202", primary: "#0ee768", secondary: "#f30404", accent: "#ffffff"},
  easter: {text: "#f7defc", background: "#09010c", primary: "#dc7df2", secondary: "#f60087", accent: "#6cdeb8"},
}
function getEaster(Y) {
  const a = Y % 19,
      b = Math.floor(Y / 100),
      c = Y % 100,
      d = Math.floor(b / 4),
      e = b % 4,
      f = Math.floor((b + 8) / 25),
      g = Math.floor((b - f + 1) / 3),
      h = (19 * a + b - d - g + 15) % 30,
      i = Math.floor(c / 4),
      k = c % 4,
      L = (32 + 2 * e + 2 * i - h - k) % 7,
      m = Math.floor((a + 11 * h + 22 * L) / 451),
      month = Math.floor((h + L - 7 * m + 114) / 31),
      day = ((h + L - 7 * m + 114) % 31) + 1;
  return new Date(Y, month - 1, day);
}
function getEvent() {
  const date = new Date();
  const easter = getEaster(date.getFullYear());
  const month = date.getMonth() + 1;
  const day = date.getDay();
  if (month == 1 && day == 1) return "new_year";
  if (month == 2 && day == 14) return "valentines_day";
  if (month == 4 && day == 1) return "april_fools";
  if (month == 10 && day == 31) return "halloween";
  if (month == 12 && day == 25) return "christmas";
  if (date.getMonth() == easter.getMonth() && date.getDate() == easter.getDate()) return "easter";
}
function setColors(colors) {
  document.body.style.setProperty("--text", colors.text);
  document.body.style.setProperty("--background", colors.background);
  document.body.style.setProperty("--primary", colors.primary);
  document.body.style.setProperty("--secondary", colors.primary);
  document.body.style.setProperty("--accent", colors.accent);
}
function update() {
  const event = getEvent();
  const c = colors[event];
  if (c) setColors(c);
  else setColors({text: null, background: null, primary: null, secondary: null, accent: null});
}
update();
setInterval(update, 1000 * 60);
})()
```
That's 1.6kb minified, or 718 bytes gzip'd. I could probably code golf it more if I wanted but I'm satisfied with this minimal implementation.

madkarmaa: so sould I just rewrite the whole lib to be functional instead of OOP?

KTibow: your original lib is nice and customizable, you shouldn't change that.
however, when we're only trying to do one thing we can take shortcuts like making it less extensible, which make it take up less size (see above)

madkarmaa: sure, I'll take care of that

oSumAtrIX: @KTibow is your intention to move the code into the website? The less on our shoulders the better. Artificial speed differences are second priority

madkarmaa: just tell me what's y'all choice and I'll implement it. I've already simplified the Easter function thanks to you and removed an entire file, which made the lib way smaller

KTibow: > The size we are talking about is less than a single image file.

Images can be lazy loaded. This needs to run *instantly*, as soon as the website loads.

> if we remove and tightly integrate into the website would stop to exist and intertwine the current codebase more than it should

I'm confused by this English. What's stopping to exist? What's "it" here?

> which made the lib way smaller

Upload to NPM so we can see it then

src/routes/+layout.svelte
madkarmaa: done in latest version of the lib

src/routes/+layout.svelte
madkarmaa: @KTibow I still don't get how to "inline" the code, can you explain it to me?

madkarmaa:
colors look a bit _odd_ but I hope they're decent. if not, they can be tweaked

package.json
```diff
@@ -41,6 +41,7 @@
 		"@tanstack/query-core": "^4.36.1",
 		"@tanstack/query-persist-client-core": "^4.36.1",
 		"@tanstack/query-sync-storage-persister": "^4.36.1",
-		"@tanstack/svelte-query": "^4.36.1"
+		"@tanstack/svelte-query": "^4.36.1",
+		"datetrigger": "^1.0.0"
```
KTibow, line 45: wouldn't be a bad idea to bump this (and to update the pnpm lock if you can)

src/app.scss
```diff
@@ -37,23 +37,23 @@ body {
...
+	--grey-three: hsl(calc(var(--hue, 206) + 15), 17%, 26%);
+	--grey-four: hsl(calc(var(--hue, 206) + 20), 48%, 18%);
+	--grey-five: hsl(calc(var(--hue, 206) + 2), 30%, 75%);
+	--grey-six: hsl(calc(var(--hue, 206) + 24), 9%, 13%);
+	--grey-seven: hsl(calc(var(--hue, 206) + 34), 9%, 13.5%);
+	--grey-eight: hsla(calc(var(--hue, 206) + 1), 30%, 75%, 0.577);
+	--grey-nine: hsla(calc(var(--hue, 206) + 34), 6%, 7%, 0.3);
+	--grey-ten: hsl(calc(var(--hue, 206) + 24), 9.5%, 17.5%);
+	--grey-eleven: hsl(calc(var(--hue, 206) + 2), 10%, 40%);
+	--red-one: hsl(calc(var(--hue, 206) + 127), 84%, 62%);
```
KTibow, line 56: i would say "red is semantic, red means danger" but apparently we only use on for the donate page? idk, it's your choice if you want to make red constant or dynamic

src/routes/+layout.svelte
```diff
@@ -42,8 +44,9 @@
 	}
 
 	onMount(() => {
-		// Check if the user has already decided.
+		new DateTriggerEventHandler(themeEvents);
 
+		// Check if the user has already decided.
```
KTibow, line 49: add a newline after this to clean up the diff

src/util/themeEvents.ts
```diff
@@ -0,0 +1,58 @@
+import { DateTriggerEvent, DAY_IN_MINUTES } from 'datetrigger';
+
+function applyCssVar(name: string, value: string): void {
+	if (!name.startsWith('--')) return console.error("A CSS variable must start with '--'");
```
KTibow, line 4:
```suggestion
	if (!name.startsWith('--')) return console.error('A CSS variable must start with "--"');
```

KTibow (commented):
we've gotten to a point where i'd call this "good enough". seemingly the two other people here don't support inlining, and the only changes aside from that are some minor ones:

package.json
madkarmaa: idk what's wrong but I can't get it to work no matter what using pnpm

src/app.scss
madkarmaa: @oSumAtrIX what's your thoughts

package.json
KTibow: `rm -rf node_modules` and then try again

src/app.scss
oSumAtrIX: I do not understand what happens here, whats the question?

madkarmaa: should the var --red-one keep the original hue or being shifted?

package.json
madkarmaa: did that multiple times, nothing changed, same error

src/app.scss
oSumAtrIX: Why are we shifting hues

madkarmaa: https://github.com/ReVanced/revanced-website/pull/227#discussion_r1499642273

oSumAtrIX: Can you elaborate what you mean?

madkarmaa: @KTibow

src/routes/+layout.svelte
madkarmaa: resolved in 56814c9

src/app.scss
KTibow: @oSumAtrIX 

Before we were hardcoding colors that we switched to for each event
Now we're hardcoding hue offsets - eg by default we have a 206 hue offset, but on easter we have a 100 hue offset. This affects most of the colors and makes them match with the scheme.
The question is whether red should always be red or change on the special events.

package.json
madkarmaa: @KTibow https://streamable.com/550jyf

Ushie:
Hey!

I would like to take a different approach than this, the shifting in the theme is very noticeable! I don't think there's any way to prevent that, so instead it would make sense to handle this at build time

We would have a theme deployer that runs on the build of the website, that creates the theme that the website will use, and we can then assign cronjobs (or something else) to redeploy the site during events automatically

I looked into what method to use to deploy the theme, whether we would manually create a file or so, but it seems like there's a package called svelte-preprocess which seems fitting

oSumAtrIX:
Build time has previously been dismissed as suitable. If the theme shifting is noticeable, add a transition if it is not preventable.

Ushie:
> Build time has previously been dismissed as unsuitable

Why? I opened all the previous reviews and I saw nothing related to this, I don't see any reason why build time would be considered unsuitable, it is the most elegant and seamless route

oSumAtrIX:
It was discussed on Discord. The reason is that it needs re-building the project.

Ushie:
Why is that a problem?

oSumAtrIX:
The final product does __not__ include the source. If your product depends on its source after building it, you are doing something wrong.

src/app.scss
oSumAtrIX: The variable red should always be red but I can imagine it being difficult for this PR to change every reference from red to some other variable. Any suggestions?

madkarmaa: resolved in 2fd62af

src/util/themeEvents.ts
```diff
@@ -0,0 +1,62 @@
+import { DateTriggerEvent, DAY_IN_MINUTES } from 'datetrigger';
+
+function applyCssVar(name: string, value: string): void {
+	if (!name.startsWith('--')) return console.error('A CSS variable must start with "--"');
+	document.documentElement.style.setProperty(name, value);
+}
```
oSumAtrIX, line 6: Should work too:
```suggestion
const val applyCssVar = document.documentElement.style.setProperty
```

src/util/themeEvents.ts
```diff
@@ -0,0 +1,62 @@
...
+		m = Math.floor((a + 11 * h + 22 * L) / 451),
+		month = Math.floor((h + L - 7 * m + 114) / 31),
+		day = ((h + L - 7 * m + 114) % 31) + 1;
+	return new Date(Y, month - 1, day);
+}
+
+const currentYear = new Date().getFullYear();
+
+export const events: DateTriggerEvent[] = [
+	// new year
```
oSumAtrIX, line 34:
```suggestion
	// New year.
```

src/util/themeEvents.ts
```diff
@@ -0,0 +1,62 @@
...
+	new DateTriggerEvent(new Date(currentYear, 0, 1), DAY_IN_MINUTES, () => {
+		applyCssVar('--hue', '240');
+	}),
+	// christmas
```
oSumAtrIX, line 38:
```suggestion
	// Christmas.
```

oSumAtrIX (commented):
Apply the comment style to the remaining comments. Apart from that LGTM

src/util/themeEvents.ts
madkarmaa: solved in bced655

src/util/themeEvents.ts
madkarmaa: solved in 4829c5a

src/util/themeEvents.ts
madkarmaa: solved in 4829c5a

src/app.scss
```diff
@@ -34,81 +38,97 @@ body {
...
-	--grey-six: hsl(230, 9%, 13%);
-	--grey-seven: hsl(240, 9%, 13.5%);
-	--grey-eight: hsla(207, 30%, 75%, 0.577);
-	--grey-nine: hsla(240, 6%, 7%, 0.3);
-	--grey-ten: hsl(230, 9.5%, 17.5%);
-	--grey-eleven: hsl(208, 10%, 40%);
+
+	--text-one: hsl(calc(var(--hue, 206) + 0), 100%, 94%);
+	--surface-one: hsl(calc(var(--hue, 206) + 0), 100%, 94%);
+
```
PalmDevs, line 46: Remove the unnecessary empty lines from here, move and order the same groups of colors together.

PalmDevs (changes requested):
All good, just need some formatting changes

PalmDevs:
Don't forget to fix conflicts! If you have issues with that, let me know so I can fix them for you.

src/app.scss
madkarmaa: the new lines exist to separate variables with the same values, making them easier to maintain

madkarmaa:
> Don't forget to fix conflicts! If you have issues with that, let me know so I can fix them for you.

please do Idk what to accept

validcube (approved):
🥞🪐 I saw the april fools preview, I'm sold.

But I think the fallback font for Comic Sans, “cursive”, should not be use because of the prominent handwritten typography, it would make the text unreadable to some. I would suggest fallback to the original font instead.

madkarmaa:
> it would make the text unreadable to some. I would suggest fallback to the original font instead.

how about a random chance of it happening? a simple page refresh would remove it. if that's still not good enough then I'll just remove it entirely

oSumAtrIX:
@madkarmaa it would be unreadable by chance then, better mske it readable for everyone

validcube:
https://github.com/ReVanced/revanced-website/commit/5b70844f3d99cf3d80e1ec01c5a6a5ffe6c8ae51 😭😭 was suggesting that we should remove the calligraphy style (“cursive”) font

oSumAtrIX:
Semantics need to be kept

PR #236, opened 2024-04-27
fix: Only show hover feedback on clickable patch list items

KTibow:
- right now, all patches look clickable, and no patches can be tabbed to.
- with this, we make expandable patches buttons (can be tabbed to), and only apply the hover animations to those. this makes it more clear when a patch can be expanded.
- we also uplevel the animation wrapper div because it was causing the widths to be odd, probably also improving perf.

### DIFF
--- a/src/routes/patches/+page.svelte
+++ b/src/routes/patches/+page.svelte
@@ -191,16 +191,14 @@
 			</PackageMenu>
 		</aside>
 
-		<div class="patches-container">
-			{#each filter(data.patches, selectedPkg || '', searchTermFiltered) as patch}
-				<!-- Trigger new animations when package or search changes (I love Svelte) -->
-				{#key selectedPkg || searchTermFiltered}
-					<div in:fly={{ y: 10, easing: quintOut, duration: 750 }}>
-						<PatchItem {patch} bind:showAllVersions />
-					</div>
-				{/key}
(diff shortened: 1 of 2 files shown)


oSumAtrIX:
Something about this feels odd. Everything has some form of feedback and you feel under control, but when hovering over the list items, it feels empty:


https://github.com/ReVanced/revanced-website/assets/13122796/2f99fa91-5daf-4ac1-8f6d-af4df0d82585

oSumAtrIX:
Discord for example gives similar visual clues even though theres no interactivity with the item itself but just some action buttons:


https://github.com/ReVanced/revanced-website/assets/13122796/c0cbc96b-02f2-4cdc-bc89-155fc69a0f4d

Ushie:
In the case of discord, there is a fork of interactivity which is the right click/context menu, but I agree that it feels empty

oSumAtrIX:
I'd argue we have interactivity with the package chip, but I can see the UX issue with showing visual cues for interaction when the item is not really interactable, but yes, it feels empty.

KTibow:
I'd argue that while removing hover effects makes it feel less interesting, it emphasizes that it's less of something to put your mouse over and more a static list. This increases semantic alignment - static elements are, well, static, and expandable items are easily noticed.

PR #239, opened 2024-04-28
fix: Remove unnecessary app version sorting

### LINKED ISSUE #238: Bug: Patches List - Sorting of YouTube's Compatible Versions (Expanded List)
Hi!.  :wave: 
And Thanks for all of Your (plural) hard work on this Awesome Project, BTW!.  :grin: 

## **issue:** 

  For some reason (that I can't seem to understand right now), when Listing the Patches for the (main) YouTube App on the Website and expanding the List of Compatible Versions, the Recommended (and, incidentally, the Latest Compatible) version of the App gets put in the middle of the List and the whole List is, weirdly, sorted in groups - that I can't understand how exactly they end up Like that / can't understand the Logic/Pattern for that Sorting.
 :thinking: 

I'm currently away from any Computer and just using Chrome on My Android Phone (and I'm in a hurry so I can't Linger on this for Long right now).

Here's a helping (shortcut) Link for the page (You still need to MANUALLY expand the List of Compatible Versions):
  **[ReVanced - Patches for ReVanced (YouTube)](https://revanced.app/patches?pkg=com.google.android.youtube)**    

Sorry that I can't help Debug and explain the reason for this big right now.  :bowing_man: 
  (I usually do...)

Here's in hopes this issue has a very easy Fix.
Cheers!.  :beers: 

  C Ya L8r then!.    :stuck_out_tongue_winking_eye:

Ushie:
I couldn't figure out why these lines of code were here, or if removing them would effect anything, but it seems to work perfectly fine
Closes #238

### DIFF
--- a/src/routes/patches/PatchItem.svelte
+++ b/src/routes/patches/PatchItem.svelte
@@ -59,8 +59,6 @@
 						const coercedA = coerce(a);
 						const coercedB = coerce(b);
 						if (coercedA && coercedB) return compare(coercedA, coercedB);
-						else if (!coercedA && !coercedB) return 0;
-						else return !coercedA ? 1 : -1;
 					})
 					.reverse() as version}
 					<li class="patch-info">
--- a/src/routes/patches/PatchItem.svelte
+++ b/src/routes/patches/PatchItem.svelte
@@ -53,14 +53,7 @@
 		<!-- Should this be hardcoded to get the version of the first package?  -->
 		{#if patch.compatiblePackages?.length && patch.compatiblePackages[0].versions?.length}
 			{#if showAllVersions}
-				{#each patch.compatiblePackages[0].versions
-					.slice()
-					.sort((a, b) => {
-						const coercedA = coerce(a);
-						const coercedB = coerce(b);
-						if (coercedA && coercedB) return compare(coercedA, coercedB);
-					})
-					.reverse() as version}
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
What do you mean? The version is not according to semver specs.

KTibow:
Yes, yet we rely on the `semver` package for sorting the items.

KTibow:
> removed

Are you saying sorting the supported versions should be entirely disabled? Or to replace it instead of entirely removing it?

oSumAtrIX:
I am saying the version should not be parsed as semve

KTibow:
Well, please explicitly state if you mean it should be parsed another way (most likely) or not parsed (literal interpretation of what you're saying)

Ushie:
How else do you propose we sort it?

oSumAtrIX:
As stated in my previous comment, there is no requirement to sort by version name at all.

KTibow:
Are you saying we should use the sorting provided by the backend instead? If so, that sounds like you should make a separate issue for that

oSumAtrIX:
The backend does not specify any sorting and is considered to be random. No issue is necessary

KTibow:
What I'm saying is if you want to remove the current sorting, you should discuss that in a place like an issue first.

oSumAtrIX:
Since it's discussed here with a clear solution, that won't be necessary anymore.

PR #251, opened 2024-08-18
fix: Use correct word break for patch option descriptions

### LINKED ISSUE #250: feat: Improve linebreak behaviour
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="assets\v2\ReVancedManager\dynamic-dark.webp">
  
</picture>

Ignore the banner, I'll make a generic one later.

# 🪄 Introduce linebreak behaviour

I suspect that the ReVanced Website doesn't use any line break behaviour for its text because of the picture below

image

## Benefits

This will greatly (even if the issue is miniscule) improve the accessibility of users using text-to-speech & readability from users with disabiliti(es) issue(s).

validcube:
fix #250

### DIFF
--- a/src/routes/patches/PatchItem.svelte
+++ b/src/routes/patches/PatchItem.svelte
@@ -110,7 +110,7 @@
 
 	#option-description {
 		white-space: pre-wrap;
-		word-break: break-all;
+		word-break: break-word;
 	}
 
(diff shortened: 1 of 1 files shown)


validcube:
wow the fix is easy, I could have done it myself.

image

Issue #191, opened 2023-10-19
feat: API status banner


Ushie:
Display a banner at the top of the webpage if the API is down/inaccessible

oSumAtrIX:
The API is inaccessible if it timeouts or returns anything other than a non-OK response. In both cases, a dedicated banner message should be shown. An offtopic idea, in addition to the API being down, can be to cache API responses in case the API is inaccessible.

Ushie:
We have this https://api.revanced.api/v2/ping endpoint which we can use, we can ping it on website initial load and check the status code

If the status code is unexpected, we can display a banner above the navbar

Ushie:
Seems like the endpoint is currently broken, will open an issue in the API repository

oSumAtrIX:
The API expects a HEAD request

Ushie:
Right

madkarmaa:
I think at the top is not the best choice, because if you add the element everything in the page is shifted downwards, and if you make the element fixed, it'll cover the navbar buttons. I think a fixed banner at the bottom of the page is ideal, even better with a close button which will keep it closed until the site is refreshed

oSumAtrIX:
If the API is down the website won't work as expected & if the banner is at the bottom it wont be noticed, especially on mobile. Additionally, eventually with announcements being shown on Website, for the latest one, we would reuse the banner to display the announcement & if the banner would be at the bottom, it would be inconsistent

PR #195, opened 2023-10-27
feat: Add announcements

oSumAtrIX:
# Todo

- [ ] Simplify access for admins through UI
- [ ] Properly implement creating announcements via the dialogue for admins
- [ ] Test the PR
- [ ] Expired image links need to be handled with an alternative properly
- [ ] Announcements can be posted to channels, channels show up at the top as filter badges, these need to be operational

PR #254, opened 2024-08-18
feat: API outage banner

madkarmaa:
# What's this?

This PR adds the `<Banner />` component and _some_ functionality. Originated from issue #191

## Screenshots (as of commit 8fa880b)

**Level: `caution`**

image

**Level: `warning`**

image

**Level: `info`**

image


oSumAtrIX:
Caution should use a different icon & all the colors are too saturated, use the same saturation/contrast as the other colors. The info banner should use the blue highlight color

madkarmaa:
> Caution should use a different icon

Such as? I've searched for "caution" on [svgrepo.com](https://www.svgrepo.com/vectors/caution/monocolor/)

> All the colors are too saturated, use the same saturation/contrast as the other colors

they're just placeholders, I didn't really follow the theme (for now)

> The info banner should use the blue highlight color

The text selection color wasn't altered, or are you referring to something else?

oSumAtrIX:
> They're just placeholders, I didn't really follow the theme (for now)

If the PR isn't ready, mark as a draft

madkarmaa:
above commit ^^^ tested to work with static build

madkarmaa:
Level: `warning`

image

Level: `caution`

image

oSumAtrIX:
Warning should be orange/yellow tinted

madkarmaa:
> Warning should be orange/yellow tinted

I couldn't find anything that follows the m3 color rules

package-lock.json
oSumAtrIX, line 1: This file should not be changed in this PR

src/app.scss
```diff
@@ -73,6 +73,14 @@ body {
...
 
 	--red-one: hsl(333, 84%, 62%);
+	--red-two: hsl(357.14, 100%, 20.59%);
+	--red-three: hsl(355.92, 100%, 28.82%);
+	--red-four: hsl(5.85, 100%, 91.96%);
+
+	--yellow-one: hsl(0, 0%, 0%);
+	--yellow-two: hsl(49.57, 100%, 4.51%);
+	--yellow-three: hsl(50.3, 39.13%, 50.39%);
+	--yellow-four: hsl(49.64, 85.94%, 74.9%);
```
oSumAtrIX, line 83: You should use colors similar to:

Red: hsl(357, 74%, 60%);
Yellow: hsl(59, 100%, 72%);

src/routes/+layout.svelte
```diff
@@ -94,6 +98,16 @@
 	</script>
 </svelte:head>
 
+{#if !data.ping.ok}
+	<Banner level="caution">
+		The API is currently down! Most of our services will be affected. Check our <a
+			href="https://status.revanced.app/"
+			target="_blank"
+			rel="noopener noreferrer">status page</a
+		> for more info.
```
oSumAtrIX, line 107:
```suggestion
		The API is currently not responding! Some services may not work correctly. Check the <a
			href="https://status.revanced.app/"
			target="_blank"
			rel="noopener noreferrer">status page</a
		> for updates.
```

src/lib/components/Banner.svelte
```diff
@@ -0,0 +1,119 @@
+<script lang="ts">
+	import { createEventDispatcher } from 'svelte';
+	import Button from './Button.svelte';
+
+	export let level: 'info' | 'warning' | 'caution' = 'info';
+
+	const dispatch = createEventDispatcher();
```
oSumAtrIX, line 7: Where is this used?

oSumAtrIX (dismissed):

src/lib/components/Banner.svelte
madkarmaa: it's used nowhere but it's functionality that might be very useful in the future

src/app.scss
madkarmaa: red is the one the m3 theme creator outputted using the default color of the website

oSumAtrIX: It's way too darky a lighter color like proposed is necessary

src/lib/components/Banner.svelte
oSumAtrIX: But why not a dismiss handler instead of a dispatcher?

madkarmaa: isn't this the intended way to handle events? the way it is now can be used this way:

```svelte
<Banner on:dismissed={someFunction} />
```

src/app.scss
madkarmaa: resolved in 79df7ad

src/routes/+layout.svelte
madkarmaa: resolved in 3e3c3ae

package-lock.json
madkarmaa: resolved in 2c4238e

src/lib/components/Banner.svelte
oSumAtrIX: I am not sure; maybe @Ushie or @PalmDevs know

src/routes/+layout.ts
```diff
@@ -1 +1,9 @@
 export const prerender = true;
+
+// https://kit.svelte.dev/docs/load#layout-data
+// https://kit.svelte.dev/docs/load#making-fetch-requests
+export async function load({ fetch }) {
```
oSumAtrIX, line 5: This function doesn't look in the right place. Shouldn't it be next to the other API queries?

src/routes/+layout.svelte
```diff
@@ -94,6 +98,16 @@
 	</script>
 </svelte:head>
 
+{#if !data.ping.ok}
+	<Banner level="caution">
+		The API is currently not responding! Some services may not work correctly. Check the <a
```
oSumAtrIX, line 103:
```suggestion
		The API is currently unresponsive, and some services may not work correctly. Check the <a
```

src/routes/+layout.ts
madkarmaa: I've linked docs, it must be there

src/lib/components/Banner.svelte
madkarmaa: it's documented in the docs (linked)

src/routes/+layout.ts
oSumAtrIX: I can not find why the load function needs to be in this file and use fetch instead of the query library like the other API calls.

madkarmaa: there was a cli warning telling me to do the `fetch` request either in the `onMount` fn or in a `load` fn. I've documented myself, and since the banner should be shown across the whole website, [this is the easiest, shortest and most readable way of doing so](https://kit.svelte.dev/docs/load#layout-data)

PalmDevs (changes requested):
I don't think we need `border-bottom` for the Banner, otherwise everything else looks good.

madkarmaa:
@PalmDevs 
> I don't think we need `border-bottom` for the Banner, otherwise everything else looks good.

solved in 079c8b8

src/routes/+layout.ts
madkarmaa: resolved in 0560498

src/routes/+layout.svelte
```diff
@@ -95,7 +100,20 @@
 </svelte:head>
 
 <QueryClientProvider client={queryClient}>
+	<Query query={pingQuery()} let:data>
+		{#if !data}
+			<Banner level="caution">
+				The API is currently unresponsive, and some services may not work correctly. Check the <a
```
oSumAtrIX, line 106:
```suggestion
				The API is currently unresponsive and some services may not work correctly. Check the <a
```

madkarmaa:
https://github.com/user-attachments/assets/2e288901-764f-4e27-9664-d153f4fea9b7

oSumAtrIX (changes requested):
When the API is down and the banner is dismissed, does it remember? Because otherwise loading another page or refreshing will cause the banner to show up again.

src/lib/components/Banner.svelte
oSumAtrIX: Why is an event dispatcher used instead of a regular lambda?

oSumAtrIX (changes requested):

src/lib/components/Banner.svelte
```diff
@@ -0,0 +1,152 @@
...
+	let banner: Element;
+
+	const removeElement = (e: AnimationEvent) => {
+		if (e.animationName.includes('swipeUp')) (e.target as Element).remove();
```
oSumAtrIX, line 11: Why is the check needed or why is the element being removed? Once the banner is outside of the view (top -100%) it isn't visible anyway

Ushie:
> When the API is down and the banner is dismissed, does it remember? Because otherwise loading another page or refreshing will cause the banner to show up again.

Refreshing the page will show it again, but going to a different page won't, I think that's good behavior because I might come back 5-10m later to check if it's working now

oSumAtrIX:
I was thinking of a permanent banner, but not as obtrusive as this one, specifically for the down API. The only reason you would want to dismiss this banner is because it blocks elements. Due to that, the user can not know if the API is up again. If the API down banner can be adjusted so that it appears as a small red strip a the top for example that can not be dismissed, but also does not block any view or space, then this problem would be solved.

The easiest way to achieve this is to add a parameter "permanent" for example. If it is true, the CSS will be used which displays the banner as a slim strip at the top of the site and hides the dismiss button. Other banners such as when used in #195 can set the parameter to false so that the banner overlaps like it currently does and has a dismiss button.

src/lib/components/Banner.svelte
madkarmaa: the check is needed because the animationend event fires for every animation, even the one that makes the banner appear

the element is removed because, since the user dismisses it, they don't want to see it again

madkarmaa:
> Refreshing the page will show it again, but going to a different page won't, I think that's good behavior because I might come back 5-10m later to check if it's working now

that's exactly what it does

src/lib/components/Banner.svelte
oSumAtrIX: If the user dismisses it the animation hides it as it is off screen, why would you need to additionally remove it?

src/lib/components/Banner.svelte
madkarmaa: Svelte official way of outputting custom events. [it can be used in plain js as well](https://developer.mozilla.org/en-US/docs/Web/Events/Creating_and_triggering_events). more readable than just accepting an inline fn from the user, because the name of the event specifically tells you when the fn you provide is gonna get fired

oSumAtrIX: What do you mean with more readable? You don't need a name for the lambda

src/lib/components/Banner.svelte
madkarmaa: why would it need to be kept in the DOM anyways?

src/lib/components/Banner.svelte
madkarmaa: can you show me an example of the way you want this to look? because [this](https://github.com/ReVanced/revanced-website/pull/254#discussion_r1727819382) is how it's currently done

src/lib/components/Banner.svelte
oSumAtrIX: It doesn't need, but it also doesn't need to be removed.

src/lib/components/Banner.svelte
oSumAtrIX: You create a variable of type lambda which is set through the component parameter and then the dismiss handler calls it.

madkarmaa: why tho, isn't `on:dismissed={ () => ... }` more readable?

oSumAtrIX: isnt { } already a lambda?

madkarmaa: https://learn.svelte.dev/tutorial/component-events

oSumAtrIX: What is this supposed to answer

src/lib/components/Banner.svelte
madkarmaa: I honestly think it does need to. imagine you have 1000 different banners, why would you leave 1000 elements in the DOM if they're not visible?

oSumAtrIX: Theres only one banner and even if it were 10 or 100 it wouldn't matter

src/lib/components/Banner.svelte
madkarmaa: { } lets you access the Event obj, it's not a lambda, it _asks_ for a lambda or a fn

oSumAtrIX: So why is an event dispatcher used when it asks for a lambda or a fn regardless?

madkarmaa: more readable? I'm following the docs, plus, if you think about it, the banner being dismissed IS an event

oSumAtrIX: How is it more readable when in both cases it asks for {fn}.

madkarmaa: `on:dismissed={() => ...}`
and (I've added a random name to the void var)
`whenDismissed={() => ...}`
are practically the same thing, the first one being how Svelte planned you to do it

madkarmaa:
@oSumAtrIX what's the practical use of a permanent banner when a simple page refresh is gonna bring it back? I don't see any user willing to stay with the website open on their browser until the API is back up again

oSumAtrIX:
The practical use is that it does not overlap over any content so the user isn't drawn to dismiss the banner & isn't needed ro refresh the page. If the API is down & the banner is dismissed there's no indication left that the site won't work properly. A permanent small banner at the top prevents that.

src/lib/components/Banner.svelte
madkarmaa: resolved in b73eb05

Ushie (changes requested):
- [x] Resolve the navbar not sticking to the top when scrolling
- [x] Confirm the API outage banner is only displayed when appropriate

Ushie (changes requested):
The `margin-top` changes across pages doesn't consider cases where there is no banner being displayed, perhaps a minor refactor where we wouldn't need magic numbers would be good

madkarmaa:
@Ushie where exactly?

oSumAtrIX:
The issue is a bit difficult to solve. The banner is sticky so we have to simulate a hidden non sticky banner so the content shifts down. I guess what we can do is add the margin to that simulated banner, so when the banner is hidden, the margin is gone too

.env
oSumAtrIX, line 1: This file should not be checked in.

src/layout/Navbar/NavHost.svelte
```diff
@@ -56,53 +61,90 @@
...
+		{#key menuOpen}
+			<div
+				id="nav-wrapper-container"
+				class:desktop-only={!menuOpen}
+				transition:horizontalSlide={{ direction: 'inline', easing: expoOut, duration: 400 }}
+			>
+				<div id="banner-pad">
+					<Query query={pingQuery()} let:data>
+						<span class="banner">
+							{#if !data}
```
oSumAtrIX, line 101: The span should be inside the if scope

static/icons/warning.svg
oSumAtrIX, line 1: Instead of the svgs, the icons package can be used once the PR is merged.

src/layout/Navbar/NavHost.svelte
```diff
@@ -56,53 +61,90 @@
...
+								<Banner level="caution" permanent>
+									The API is currently unresponsive and some services may not work correctly. Check
+									the <a
+										href="https://status.revanced.app/"
```
oSumAtrIX, line 105: Hardcoding the status page may not be the best idea, but there is no other place that links to it. It'll be present in https://api.revanced.app/v4/about under the key "status" in the next API release.

Ushie: If the API is down, how is the website expected to parse that from the API?

src/layout/Navbar/NavHost.svelte
Ushie: It's needed for styling purposes and span is being used for its intended purpose

src/layout/Navbar/NavHost.svelte
oSumAtrIX: In this case the status page can be a compile time variable just like the API url.

oSumAtrIX:
No banner is shown despite the API being unreachable:

image

image

PR #257, opened 2024-09-08
feat: Add announcements

madkarmaa:
## What's this?
This is a new, more updated continuation of #195

### TODO:

- [x] Show banner for latest unread announcement (https://github.com/ReVanced/revanced-website/pull/254)


Ushie:
> Only first attachment shown on announcements page

Well, that's intentional, as it's a thumbnail not a gallery

oSumAtrIX:
The first attachment is not intended to be a thumbnail. There is no concept of thumbnails in announcements, so you arbitrarily chose the first as one. Can't say if this is really a good idea

madkarmaa:
image

Ushie:
> In mobile view the separators should be full width

Without making the CSS horrible, the padding doesn't really allow for that

Ushie:
> Inspecting the network tab, query is making fetches every couple of seconds for individual announcements for no reason

I'm not able to recreate this

oSumAtrIX:
Click on some announcements. Query will try to refresh the value from time to time.

src/layout/Navbar/NavHost.svelte
```diff
@@ -132,12 +165,144 @@
...
+		</div>
+	</svelte:fragment>
+</Modal>
+
+<!-- login -->
+<Modal bind:modalOpen={loginOpen}>
+	<div class="admin-modal-content">
+		<h2>Admin Login</h2>
+		<p>
+			This panel is reserved for administrators at ReVanced, this is not what you should be looking
```
oSumAtrIX, line 189:
```suggestion
			This login is reserved for site administrators.
```

src/layout/Navbar/NavHost.svelte
```diff
@@ -132,12 +165,144 @@
...
+				<Button type="text" on:click={save} label="Save Button">Save</Button>
+			</div>
+		</div>
+	</svelte:fragment>
+</Modal>
+
+<!-- login -->
+<Modal bind:modalOpen={loginOpen}>
+	<div class="admin-modal-content">
+		<h2>Admin Login</h2>
```
oSumAtrIX, line 187:
```suggestion
		<h2>Login</h2>
```

Ushie:
> Deleting tags does not behave like when creating a new announcement, they should disappear once unchecked

It has nothing to do with whether or not you're creating an announcement

Tags that already exist in the API will not disappear, tags that you've created in that draft will disappear once unchecked

Ushie:
> Unread latest announcement should be shown as a banner on the website.

The banners PR is not complete

Ushie:
> Nullability is not handled properly

ReVanced API seems to also not respect it

Creation:
image

Result:
image

oSumAtrIX:
> > Deleting tags does not behave like when creating a new announcement, they should disappear once unchecked
> 
> It has nothing to do with whether or not you're creating an announcement
> 
> Tags that already exist in the API will not disappear, tags that you've created in that draft will disappear once unchecked

Even if they don't disappear in the API, they do for the announcement itself. So if you uncheck a tag, it should be removed. Adding it back via the same name, would not create a new tag because it already exists. This behaviour is the same as on YouTube if you tag videos.


> > Unread latest announcement should be shown as a banner on the website.
> 
> The banners PR is not complete

In that case it should be looked into so that this PR can rebase onto it.

> ReVanced API seems to also not respect it

The tags and attachments fields are non nullable. However I'll change that. For author the issue still applies.

PR #271, opened 2024-12-13
feat: Use icon library instead of SVGs

### LINKED ISSUE #255: Issues with icons
I've come across some issues while using the `<Button />` component.

- There's a limited choice of icons, which need to be added manually

- The colors of the icons are hardcoded in the SVG code

[I had to modify the `filter` CSS property to make the colors of the icons black/white](https://github.com/ReVanced/revanced-website/pull/254/files#diff-591f218295070c51913726fb03f7dc03f3edb3224c3fff6cfd2ff0481f2434ccR95-R99) because the hardcoded blue-ish color they currently have didn't fit on red/yellow backgrounds.

I think having an icon library is better than the current method, such as [Google font icons](https://fonts.google.com/icons) or [FontAwesome icons](https://fontawesome.com/icons).

A component can be created as well, [I've already made one for my own website](https://github.com/madkarmaa/madkarmaa.github.io/blob/master/src%2Flib%2Fcomponents%2FFAIcon.svelte), it's a really easy solution.

The only downside I can think of is the big size of these CSS files which need to be downloaded by the client.

madkarmaa:
## What's this?

A simple component that replaces the use of SVGs in the website.

closes #255


oSumAtrIX (approved):
Theres some more resources in the static dir for icons iirc. Should they be possible to replace, please do so in this PR as well.

madkarmaa:
> Theres some more resources in the static dir for icons iirc. Should they be possible to replace, please do so in this PR as well.

can you explain this further?

oSumAtrIX:
See https://github.com/ReVanced/revanced-website/tree/main/static/icons

madkarmaa:
> See https://github.com/ReVanced/revanced-website/tree/main/static/icons

everything should be swapped out now

src/lib/components/Icon.svelte
oSumAtrIX, line 1: This component can be removed.

oSumAtrIX (approved):
a container for the icon shouldnt be necessary. If you need to change the icon directly, use the components props. If you need to align it or similar, change the outer div.

package.json
```diff
@@ -17,6 +17,7 @@
 		"@sveltejs/adapter-static": "3.0.6",
 		"@sveltejs/kit": "2.9.0",
 		"@sveltejs/vite-plugin-svelte": "3.1.1",
+		"@types/semver": "^7.5.8",
```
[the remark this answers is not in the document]
madkarmaa: only the types. semver was already present. it is installed but not used, and the ide was complaining about the library missing types

src/layout/Hero/HeroSection.svelte
```diff
@@ -9,8 +12,18 @@
...
+					<TrayArrowDown size="20px" />
+				</svelte:fragment>
+				Download
+			</Button>
+			<Button type="tonal" href="patches">
+				<svelte:fragment slot="icon">
+					<FileDocumentOutline size="20px" />
+				</svelte:fragment>
+				View patches
+			</Button>
```
Ushie, line 26: Why the fragments? can't the icons be included with the text?

madkarmaa: makes it more obvious what it is?

Ushie: One of the only benefits of this PR is improving the DX when it comes to using icons, because inlining SVGs takes up too much screen estate and importing them doesn't allow them to be colored dynamically (without weird hacks)

This isn't much better if it's going to take 3 lines to display the icon, why not make the icon a parameter to the Button component or just inline it with the text?

Ushie (approved):

PR #273, opened 2025-01-08
feat: Add the footer to the homepage

Ushie:


oSumAtrIX:
Seems like something wen't wrong here:

https://github.com/user-attachments/assets/40aab2b7-ef77-4fd5-b9a0-8dad001f4bff

oSumAtrIX:
The socials are missing as well:

image

Ushie:
The socials are missing because the API blocks the local dev server through CORS

PR #294, opened 2025-05-30
fix: Only render GPG key tooltip and icon if team member has one

fedeericodl:
Website is crashing right now if the `/donate` page is viewed, due to a team member not having a GPG key. This will render the GPG checkmark icon and tooltip only if the team member has one set.
It is useful to have this check in case it happens again in the future for even longer periods maybe.

### DIFF
--- a/src/routes/donate/TeamMember.svelte
+++ b/src/routes/donate/TeamMember.svelte
@@ -28,21 +28,23 @@
 				<h4>{member.name}</h4>
 			</a>
 			<div class="verified-badge">
-				<ToolTip
-					content="<p>GPG key ID:</p> <a class='gpg-url' href={member.gpg_key
-						.url} rel='noreferrer' target='_blank'>{member.gpg_key.id}</a>"
-					html={true}
-				>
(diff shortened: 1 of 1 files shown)


src/routes/donate/TeamMember.svelte
```diff
@@ -28,21 +28,23 @@
...
 			</a>
 			<div class="verified-badge">
-				<ToolTip
-					content="<p>GPG key ID:</p> <a class='gpg-url' href={member.gpg_key
-						.url} rel='noreferrer' target='_blank'>{member.gpg_key.id}</a>"
-					html={true}
-				>
-					<div class="desktop">
-						<a href={member.gpg_key.url} rel="noreferrer" target="_blank">
+				{#if member.gpg_key}
```
oSumAtrIX, line 31: The badge would still show up even without a gpg key. This is not correct behavior

fedeericodl: Oops. My bad, was an oversight. Now it should be correct, I should have grouped the entire badge element.

PR #318, opened 2025-07-08
ci: Install the exact same dependencies as lockfile 

validcube:


.github/workflows/deploy.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           cache: npm
 
       - name: Install dependencies
-        run: npm i
+        run: npm ci
```
oSumAtrIX, line 30: I just checked and it does:

```
If a node_modules is already present, it will be automatically removed before npm ci begins its install.
```

How does this affect the NPM cache in this workflow? Does it not render it useless?

PR #327, opened 2025-09-16
feat: Add notice before showing e-mail

oSumAtrIX:
# Todo

- Move to API/env var
- Add a new mail specific to the site in API (e.g. nosupport@revanced.app)


Ushie:
Are both todos not the same?

Ushie (approved):

PR #352, opened 2026-02-01
fix: fix announcements creation

madkarmaa:
`created_at` was being passed in the request body, and the request was getting rejected by the api as `created_at` should only be defined by the backend


oSumAtrIX:
Thats how it should be working in the api, if you dont set created at, the api chooses now, if you specify, it will respect that

PR #392, opened 2026-05-06
Exclude more bots in the contributor page

SayanthD:


src/lib/components/organisms/ContributorSection.svelte
```diff
@@ -15,7 +15,7 @@
 
 	let expanded = $state(true);
 
-	const botNames = ['semantic-release-bot', 'revanced-bot'];
+	const botNames = ['semantic-release-bot', 'revanced-bot', 'dependabot[bot]', 'github-actions[bot]', 'pre-commit-ci[bot]'];
```
oSumAtrIX, line 18: Maybe this should be handled at the API layer in revanced-api?

SayanthD: I would prefer to filter this on the frontend. It’s better to keep what the API returns as it is.

oSumAtrIX: The API is also used by revanced manager and other frontends, why would manager for example not be subject of this filter?

SayanthD: Yep Just saw that in the Manager. It should be filtered as well

SayanthD:
Closing as https://github.com/ReVanced/revanced-api/pull/223 will supersede this once merged

PR #393, opened 2026-05-18
feat: Implement a fallback system to the ReVanced Website

mostafaNazari702:
This pull request implements a fallback system for the ReVanced Website should the primary and fallback API not be available, an offline/cache-state will be implemented to keep the website operational even if:
1. The user is offline but has, from earlier visits, cached the website data
2. The primary API and the fallback API are unresponsive 

The process will be iterating rather than one massive PR. The maintainer(s) can suggest or push back on anything they see and their opinions matter on each PUSH, "continue" is required for me to continue with the next part. A maximum of 2 commits will be pushed a push.


src/lib/api/server.ts
```diff
@@ -18,19 +18,96 @@ import {
 import type { z } from 'zod';
 
 const API_VERSION = 'v5';
+const DEFAULT_CACHE_TTL_SECONDS = 300;
+const CACHE_STORED_AT_HEADER = 'x-rv-cache-stored-at';
```
oSumAtrIX: Whats this? No such header exists

src/lib/api/server.ts
```diff
@@ -18,19 +18,96 @@ import {
...
+const CACHE_TTL_HEADER = 'x-rv-cache-ttl';
```
oSumAtrIX: there is an official cache-control header, dont use a custom header

src/lib/api/server.ts
```diff
@@ -18,19 +18,96 @@ import {
...
+	const cacheable = new Response(body, {
+		status: response.status,
+		statusText: response.statusText,
+		headers
+	});
+	await cache.put(cacheKey, cacheable);
+}
+
+async function fetchWithEdgeCache(url: string, fetchFn: typeof fetch): Promise<Response> {
+	const cache = getEdgeCache();
```
oSumAtrIX: Whats this? Browsers already have a cache

src/lib/api/server.ts
```diff
@@ -18,19 +18,96 @@ import {
...
+
+	try {
+		const fresh = await fetchFn(url);
+		if (fresh.ok) {
+			const { maxAge, noStore, noCache } = parseCacheControl(fresh);
+			if (!noStore) {
+				const ttl = noCache ? 0 : (maxAge ?? DEFAULT_CACHE_TTL_SECONDS);
+				await storeInEdgeCache(cache, cacheKey, fresh, ttl);
+			}
+			return fresh;
```
oSumAtrIX: Browsers already handle cache control automatically

.env.example
```diff
@@ -1,4 +1,5 @@
 RV_API_URL=https://api.revanced.app
+RV_API_URL_FALLBACK=
```
oSumAtrIX, line 2: By default, I suggest using my personal domain. The reason is that my domain is predictably live, and I am in a trustworthy position, which is better than pointing to a non-existing/dormant domain that has to be purchased/maintained offline. The fallback will act for emergency and potential migration to a new suitable domain if necessary. Since this .env.example file is mimicking our own deployment for demonstrative purposes, this default domain is added here. This env var is only used as a default value for the website, in the instance, it never manages to contact the api once to override this value.  

```suggestion
RV_API_URL_FALLBACK=https://revanced.osumatrix.me
```

Open to opinions @Ushie @Axelen123.

oSumAtrIX (commented):
One consideration about the fallback system is if we want to have a capture feature where a fallback can prevent the site recovering to the original URL, this would prevent a scenario when the main domain is taken over and brought back live. The fallback API field can be:

fallback: { url: "", recover: true }

If recover is set to false, it won't attempt to recover until the recover variable is set to true again.

oSumAtrIX (changes requested):
The fallback logic isnt implemented yet

.env.example
Ushie: Your domain can be used, that's an infrastructure matter, but it doesn't need to be set as the default within the example, by default it should remain blank so invalid APIs is treated properly in local dev envs instead of it falling back to your domain aswell which would be unpredictable in development

Fallbacks only matter for production

oSumAtrIX: The .env vars mimick our current deployements as best as possible so you have a full reference of a .env var:

For example here our contact email is set for examplary purposes:
https://github.com/ReVanced/revanced-website/blob/main/.env.example

Here our repos are used:
https://github.com/ReVanced/revanced-api/blob/main/.env.example

Someone setting .env up would want to change them

Ushie: Would want to change them yes, but most of the time .env.example are preconfigured as a development optimal starting point in my experience, where for something simple I should be able to just remove the .example extension and start development

besides there is no motivation to point to outside the revanced domain officially in source

oSumAtrIX: Our deployment is what is considered "optimal". What we deploy, we deem as the most optimal. In this case, the fallback is a mechanism that must link out of revanceds domain. Its purpose is to recover the client if the domain is down, ideally somewhere out of the collateral zone (similar to when you setup a recovery account for your google account which notifies someone when your account is inactive). The fallback url field is nullable so technically a null .env is valid, however the api is considered to be deployed as the main one while fallbacks are considered secondary which is why an .env var is set here

mostafaNazari702:
I don't think ssr/worker sees localstorage so the "recover: false" capture is on the browser only i think. it continues following RV_API_URL => RV_API_URL_FALLBACK (env only). full server-side capture needs persisting the dynamic value in a DO, it will be a follow-up. Probably after this.

src/lib/api/client.ts
```diff
@@ -31,15 +31,36 @@ class ApiValidationError extends Error {
...
+	throw lastError;
+}
+
+async function resilientFetch(endpoint: string, init?: RequestInit): Promise<Response> {
+	try {
+		return await tryWithRetries(composeApiUrl(getApiUrl(), endpoint), init);
+	} catch (primaryErr) {
+		const fallback = getFallbackApiUrl();
+		if (!fallback) throw primaryErr;
+		return tryWithRetries(composeApiUrl(fallback, endpoint), init);
```
oSumAtrIX: When the fallback specifies another fallback, it wont try it. You probbaly want to call recursively until there is no fallback specified

src/lib/api/server.ts
```diff
@@ -15,22 +15,96 @@ import {
...
+	if (/\bmax-age\b/i.test(cc)) {
+		await cache.put(cacheKey, fresh.clone());
+	} else {
+		const body = await fresh.clone().arrayBuffer();
+		const headers = new Headers(fresh.headers);
+		headers.set('Cache-Control', `public, max-age=${DEFAULT_MAX_AGE_SECONDS}`);
+		await cache.put(
+			cacheKey,
+			new Response(body, { status: fresh.status, statusText: fresh.statusText, headers })
+		);
```
oSumAtrIX: This is very raw, use the dedicated API:

https://developers.cloudflare.com/workers/examples/cache-using-fetch/

This is cloudflare specific though, you might want to make this adapter specific.

src/lib/api/server.ts
```diff
@@ -15,22 +15,96 @@ import {
...
+	if (stored && stored.recover === false && stored.url) {
+		return { primary: stored.url, fallback: null };
+	}
+	return { primary: RV_API_URL, fallback: stored?.url || envFallback };
+}
+
+async function resilientFetch(endpoint: string, fetchFn: typeof fetch): Promise<Response> {
+	const { primary, fallback } = await getServerActiveUrls();
+	try {
+		return await fetchWithRetries(composeApiUrl(primary, endpoint), fetchFn);
```
oSumAtrIX: Same as the client side, you may want to do it recursively

src/lib/api/storage.ts
```diff
@@ -0,0 +1,48 @@
+import { AsyncLocalStorage } from 'node:async_hooks';
+
+export interface Storage {
+	get(key: string): Promise<string | null>;
+	set(key: string, value: string): Promise<void>;
+	delete(key: string): Promise<void>;
+}
+
+type CloudflareKV = {
```
oSumAtrIX, line 6: Any reason durable objects were not used here

src/hooks.server.ts
```diff
@@ -0,0 +1,7 @@
+import type { Handle } from '@sveltejs/kit';
+import { createStorageFromPlatform, runWithStorage } from '$lib/api/storage';
+
+export const handle: Handle = async ({ event, resolve }) => {
+	const storage = createStorageFromPlatform(event.platform);
+	return runWithStorage(storage, () => resolve(event));
+};
```
oSumAtrIX: Whats this for

src/lib/stores/readAnnouncements.svelte.ts
```diff
@@ -1,56 +1,13 @@
...
-
-		return right.id - left.id;
-	});
-}
 
 function getHighestAnnouncementId(announcements: Array<{ id: number }>): number {
 	return announcements.reduce((highestId, announcement) => Math.max(highestId, announcement.id), 0);
 }
 
-function getLatestAnnouncement(
```
oSumAtrIX, line 27: If this is removed, how do you check if there is a new latest announcement to show the banner

src/lib/components/organisms/AppShell.svelte
```diff
@@ -37,12 +37,10 @@
 	let contactEmail = $derived(about?.contact?.email ?? '');
 	let apiIsDown = $derived(about === null);
 	let bannerAnnouncements = $derived(
-		announcementPolling.data.length > 0
-			? announcementPolling.data
-			: latestAnnouncements
-					.map(({ announcement }) => announcement)
-					.sort((left, right) => right.id - left.id)
```
oSumAtrIX, line 44: There is a /latest api, are you using that? https://api.revanced.app/v5/announcements/latest

src/lib/api/client.ts
mostafaNazari702: I just want to make sure we're talking about the same behavior before I change this.

Option A is that within a single request, if URL X fails we fetch /about from its fallback Y, then if needed fetch /about from Y's fallback Z, and keep going until there are no more fallbacks. The thing is is that every failover adds another /about request. Option B is more incremental, whenever a client successfully connects to URL X, we store X => X.fallback in localStorage. Then on future failovers we walk that cached map instead. in that model the chain gets built up over time rather than during a single request.

for now the /about response only supports a single fallback (`fallback: { url, recover }`) so deeper chains can only be discovered by walking them step by step.

which approach were you thinking of? or would you rather change /about so it returns the full fallback chain in one response, like `fallback: [{ url, recover }, ......]` ?

src/lib/api/storage.ts
mostafaNazari702: KV was mostly just the practical choice with the current setup.
rn this repo uses @sveltejs/adapter-cloudflare with pages_build_output_dir so it's running through the Pages adapter. with that setup Durable Object classes can't be in the same worker as the sveltekit handler. they need to be defined in a separate Worker deployed separately and then linked back in via script_name in wrangler.toml.

KV on the other hand is basically just a single binding and works without any extra deployment setup.

the storage layer is already abtracted behind the Storage interface so switching to DO later would be pretty small. It'd just mean adding something like a durableObjectStorage implements Storage class and updating the binding in wrangler.toml ig.

So: that route, separate worker setup or KV since we're really only storing a single object..?

src/lib/components/organisms/AppShell.svelte
mostafaNazari702: In short: Yes

But longer:
in src/lib/api/server.ts it calls the announcements/latest endpoint and then +layout.server.ts => AppShell ( as the latestAnnouncements prop ). AppShell.svelte then uses that prop for the banner logic.

src/lib/api/client.ts
oSumAtrIX: > Option B is more incremental, whenever a client successfully connects to URL X, we store X => X.fallback in localStorage. Then on future failovers we walk that cached map instead. in that model the chain gets built up over time rather than during a single request.

X.fallback might have another fallback, so recursion is still needed.

> which approach were you thinking of

If the client builds the fallback chain and caches it, the cache might be not up to date, i'd say option a is better, the fallback will rarely be used, and if, then one longer request is fine.

src/lib/api/storage.ts
oSumAtrIX: KV is the correct solution for when you need a list of entries, but here we just save one object which makes durable objects the correct solution.

I am not aware of how cloudflare adapter handles this here, but just like kv, i believe durable objects are just bindings too. Durable objects are obviously only writeable in a server.ts, so serverside, but i dont see any problems with that. The cloudflare adapter already needs a worker.toml as far as i know.

src/lib/components/organisms/AppShell.svelte
oSumAtrIX: yes thats correct, the api call is made by the websites server, and renders the client, server side, cached by the frontend proxy (in our case cf)

src/lib/api/client.ts
mostafaNazari702: > > Option B is more incremental, whenever a client successfully connects to URL X, we store X => X.fallback in localStorage. Then on future failovers we walk that cached map instead. in that model the chain gets built up over time rather than during a single request.
> 
> X.fallback might have another fallback, so recursion is still needed.
> 
> > which approach were you thinking of
> 
> If the client builds the fallback chain and caches it, the cache might be not up to date, i'd say option a is better, the fallback will rarely be used, and if, then one longer request is fine.

I think you are a bit off when it comes to deplyment mechanic, Pages adapter @sveltejs/adapter-cloudflare +
pages_build_output_dir can't define a DO class inside the SvelteKit-generated worker, Pages binds to DOs but the class itself has to live in a separately-deployed Workers script and be referenced via `script_name`. so shipping DO means two deployments: the Pages site + a small standalone Worker that exports the `FallbackStorage` class. Wrangler binding would look like:

```toml
[[durable_objects.bindings]]
name = "FALLBACK_STORAGE"
class_name = "FallbackStorage"
script_name = "revanced-website-storage"
```

oSumAtrIX: please see https://github.com/sveltejs/kit/issues/1712 and https://www.npmjs.com/package/@omysoul/cf-ws-do

mostafaNazari702: > please see [sveltejs/kit#1712](https://github.com/sveltejs/kit/issues/1712) and https://www.npmjs.com/package/@omysoul/cf-ws-do

I have gone through all of those and checked and hopefully did not miss anything important to learn about.
Right now, the official adapter still doesn't support inlining durable objects in either mode. sveltejs/kit#1712 was closed without that landing, and cloudflare's own Pages docs explicitly say: "You cannot create and deploy a Durable Object within a Pages project."

So realistically here and for your needs counted in, the best option seems to use @omysoul/cf-ws-do ( if you are not already implying to use that ):
- moves this repo from Pages to Workers Static Assets
- uses an alpha community adapter
- single deployment setup
- DOs live directly in *.do.ts files

Not sure about that but probably the cleaner long-term setup BUT it does mean switching away from the official adapter and changing the deployment model.
Should i go ahead with that or do you have other input?

mostafaNazari702: @oSumAtrIX I will go ahead with that in a moment, if you have any comments please express.

oSumAtrIX: It depends on how major the changes are, if it's just minor changes without any issues, sure, otherwise, can't you manually call the cf Apis to access durable objects?

mostafaNazari702: > It depends on how major the changes are, if it's just minor changes without any issues, sure, otherwise, can't you manually call the cf Apis to access durable objects?

I think i will go the "manually call the cf Apis" route.

oSumAtrIX: Yes, or maybe you can inject the bindings manually, I'm not sure how it works exactly

mostafaNazari702:
revanced-website-storage is separate cloudflare worker for the Durable Object storage for /about.fallback. 
u will have to deploy it on on its own and link it through script_name in wrangler.toml.

__After the first deploy__ uncomment the Durable Object binding in the main wrangler.toml and redeploy the Pages app.
Probably should be in its own but it is to little so i guess it is cool to keep it there.

src/lib/api/client.ts
```diff
@@ -31,15 +31,36 @@ class ApiValidationError extends Error {
...
+		try {
+			return await fetch(url, init);
+		} catch (err) {
+			lastError = err;
+		}
+	}
+	throw lastError;
+}
+
+async function resilientFetch(endpoint: string, init?: RequestInit): Promise<Response> {
```
oSumAtrIX: rename to fetchWithFallback

src/lib/api/server.ts
```diff
@@ -15,22 +15,102 @@ import {
...
-const API_VERSION = 'v5';
+const DEFAULT_MAX_AGE_SECONDS = 300;
+const MAX_RETRIES = 3;
+const FALLBACK_STORAGE_KEY = 'fallback';
 
-function buildServerUrl(endpoint: string): string {
-	endpoint = endpoint.replace(/^\/+/, '');
-	return `${RV_API_URL}/${API_VERSION}/${endpoint}`;
+async function fetchWithEdgeCache(url: string, fetchFn: typeof fetch): Promise<Response> {
+	const cache = (globalThis as { caches?: { default?: Cache } }).caches?.default;
```
oSumAtrIX: Why is this needed?

src/lib/api/server.ts
```diff
@@ -15,22 +15,102 @@ import {
...
+	if (!cache) return fetchFn(url);
+
+	const cacheKey = new Request(url);
```
oSumAtrIX: Does the key depend on headers too? A request could have a datetime in it and then every request would be different and never match. Instead the key can simply be the URL and http method

src/lib/api/server.ts
```diff
@@ -15,22 +15,102 @@ import {
...
+	const cached = await cache.match(cacheKey);
+	if (cached) return cached;
+
+	const fresh = await fetchFn(url);
+	if (!fresh.ok) return fresh;
```
oSumAtrIX: Just do 

if (cache) {
 const cached = await cache.match(cacheKey)
...
}

src/lib/api/server.ts
```diff
@@ -15,22 +15,102 @@ import {
...
+
+	const cc = fresh.headers.get('Cache-Control')?.toLowerCase() ?? '';
```
oSumAtrIX: dont shorten like this, call it cacheControl

src/lib/api/server.ts
```diff
@@ -15,22 +15,102 @@ import {
...
+	if (/\bno-store\b|\bno-cache\b|\bprivate\b/.test(cc)) return fresh;
+
+	if (/\bmax-age\b/.test(cc)) {
+		await cache.put(cacheKey, fresh.clone());
```
oSumAtrIX: This is manual and hacky, isnt there a library for this? Cloudflare workers fetch already has native handling of caching.

oSumAtrIX: I suppose due to the 5min fallback?

oSumAtrIX: I think we can use cloudflare official fetch caching feature. Simply do:

if (cacheControl header missing) request.header[Cache control] = 5 min.

This way we can utilize fetch'es native cache feature with a fallback to 5min if no cache header is present. 

Another thing is to abstract the persistence layer and the http client. Both hardcode cloudflare native APIs but should be abstracted behind an interface so in case we want to swap fetch implementation and the persistence layer for caching, we can simply swap the interface.

src/lib/api/server.ts
```diff
@@ -15,22 +15,102 @@ import {
...
+		headers.set('Cache-Control', `public, max-age=${DEFAULT_MAX_AGE_SECONDS}`);
+		await cache.put(
+			cacheKey,
+			new Response(body, { status: fresh.status, statusText: fresh.statusText, headers })
+		);
+	}
+	return fresh;
+}
+
+async function fetchWithRetries(url: string, fetchFn: typeof fetch): Promise<Response> {
```
oSumAtrIX: why is client api called tryWithRetries and this fetchWithRetries? Cant you share the code between server and client?

src/lib/api/server.ts
```diff
@@ -15,22 +15,102 @@ import {
...
+	let stored: { url: string | null; recover: boolean } | null = null;
+	try {
+		const raw = await storage.get(FALLBACK_STORAGE_KEY);
+		if (raw) {
+			const parsed = JSON.parse(raw);
+			if (typeof parsed === 'object' && parsed !== null) {
+				stored = {
+					url: typeof parsed.url === 'string' ? parsed.url : null,
+					recover: parsed.recover !== false
+				};
```
oSumAtrIX: Way too complex, just parse the object and cast to a type

src/lib/stores/readAnnouncements.svelte.ts
```diff
@@ -1,54 +1,67 @@
...
+	onUpdate: (data: TaggedLatestAnnouncements[]) => void
+): () => void {
+	if (!browser) return () => {};
+
+	let timer: ReturnType<typeof setTimeout> | null = null;
+	let canceled = false;
 
-	for (const announcement of latestAnnouncements) {
-		byId.set(announcement.id, announcement);
+	async function poll(): Promise<void> {
```
oSumAtrIX: Why did you add polling and client side fetching? The website is intentionally server side rendered, so this shouldnt be needed

src/app.d.ts
```diff
@@ -7,7 +7,15 @@ declare global {
 		// interface Locals {}
 		// interface PageData {}
 		// interface PageState {}
-		// interface Platform {}
+		interface Platform {
+			env?: {
+				FALLBACK_STORAGE?: {
```
oSumAtrIX: This should simply be called storage. Why is it placed in Platform.env?

wrangler.toml
```diff
@@ -2,3 +2,13 @@ name = "revanced-website"
...
 pages_build_output_dir = ".svelte-kit/cloudflare"
 compatibility_flags = ["nodejs_compat"]
+
+# server-side persistence for the fallback URL from the /about endpoint.
+# We need this for the "recover: false" capture mechanism to also bite at the SSR layer (without it capture is browser-only)
+# Pages cant host Durable Object classes and we agreed not to deploy a separate worker just for this so KV is the persistence layer.
+# the Storage interface in src/lib/api/storage.ts keeps the impl swappable if we ever migrate to DO.
+# Create the namespace with "wrangler kv namespace create FALLBACK_STORAGE", paste the returned id below, then uncomment.
+# until then the SSR path falls through to env-only behavior.
+# [[kv_namespaces]]
```
oSumAtrIX: So weird that sveltekit supports this, but not durable objects

wrangler.toml
```diff
@@ -2,3 +2,13 @@ name = "revanced-website"
 compatibility_date = "2026-03-14"
 pages_build_output_dir = ".svelte-kit/cloudflare"
 compatibility_flags = ["nodejs_compat"]
+
+# server-side persistence for the fallback URL from the /about endpoint.
+# We need this for the "recover: false" capture mechanism to also bite at the SSR layer (without it capture is browser-only)
+# Pages cant host Durable Object classes and we agreed not to deploy a separate worker just for this so KV is the persistence layer.
+# the Storage interface in src/lib/api/storage.ts keeps the impl swappable if we ever migrate to DO.
```
oSumAtrIX:
```suggestion
# the Storage interface in src/lib/api/storage.ts keeps the implementation swappable once DO is supported.
```

wrangler.toml
```diff
@@ -2,3 +2,13 @@ name = "revanced-website"
 compatibility_date = "2026-03-14"
 pages_build_output_dir = ".svelte-kit/cloudflare"
 compatibility_flags = ["nodejs_compat"]
+
+# server-side persistence for the fallback URL from the /about endpoint.
+# We need this for the "recover: false" capture mechanism to also bite at the SSR layer (without it capture is browser-only)
+# Pages cant host Durable Object classes and we agreed not to deploy a separate worker just for this so KV is the persistence layer.
```
oSumAtrIX: Pages can host DO, what cant is sveltekit


```suggestion
# SvelteKit does not yet support Durable Objects and we agreed not to deploy a separate worker just for this so KV is the persistence layer.
```

wrangler.toml
```diff
@@ -2,3 +2,13 @@ name = "revanced-website"
 compatibility_date = "2026-03-14"
 pages_build_output_dir = ".svelte-kit/cloudflare"
 compatibility_flags = ["nodejs_compat"]
+
+# server-side persistence for the fallback URL from the /about endpoint.
```
oSumAtrIX:
```suggestion
# Server-side persistence for the fallback URL from the /about endpoint.
```

wrangler.toml
```diff
@@ -2,3 +2,13 @@ name = "revanced-website"
...
+# We need this for the "recover: false" capture mechanism to also bite at the SSR layer (without it capture is browser-only)
```
oSumAtrIX: Try to reformulate comments. Instead of saying "we" say something else like. ""recover: false" is needed...

wrangler.toml
```diff
@@ -2,3 +2,13 @@ name = "revanced-website"
...
+# Pages cant host Durable Object classes and we agreed not to deploy a separate worker just for this so KV is the persistence layer.
+# the Storage interface in src/lib/api/storage.ts keeps the impl swappable if we ever migrate to DO.
+# Create the namespace with "wrangler kv namespace create FALLBACK_STORAGE", paste the returned id below, then uncomment.
+# until then the SSR path falls through to env-only behavior.
+# [[kv_namespaces]]
+# binding = "FALLBACK_STORAGE"
+# id = ""
```
oSumAtrIX: Please also adjust CI so at deployment it injects the id. The action should be able to, if not let me know

src/lib/api/settings.ts
```diff
@@ -9,41 +10,115 @@ import {
...
+	} else if (aboutData.fallback !== undefined) {
+		writeStoredFallback({
+			url: aboutData.fallback.url ?? null,
+			recover: aboutData.fallback.recover
+		});
 	}
 }
 
+export function getApiUrl(): string {
+	return getActiveUrls().primary;
```
oSumAtrIX: This is inefficient. Every call to getApiUrl is going to make a read to the persistence layer. The worker does multiple fetches. Please use constants wherever possible and reduce unnecessary calls

src/lib/api/settings.ts
```diff
@@ -9,41 +10,115 @@ import {
...
-		const cached = localStorage.getItem(STATUS_KEY);
-		if (cached) return cached;
-	}
-	return DEFAULT_STATUS_URL;
+	return readLocal(STATUS_KEY) ?? DEFAULT_STATUS_URL;
 }
 
 export function getContactEmail(): string {
-	if (browser) {
-		const cached = localStorage.getItem(EMAIL_KEY);
```
oSumAtrIX, line 43: Why was the email being cached in local storage?

src/lib/api/settings.ts
```diff
@@ -56,11 +131,14 @@ export function clearCacheAndReload(): void {
 	location.reload();
 }
 
-export function buildUrl(endpoint: string): string {
+export function composeApiUrl(base: string, endpoint: string): string {
 	endpoint = endpoint.replace(/^\/+/, '');
```
oSumAtrIX: Why is this needed?

src/lib/api/settings.ts
```diff
@@ -56,11 +131,14 @@ export function clearCacheAndReload(): void {
...
-	if (endpoint.startsWith(API_VERSION)) {
-		endpoint = endpoint.split('/').slice(1).join('/');
+	if (endpoint === API_VERSION || endpoint.startsWith(API_VERSION + '/')) {
+		endpoint = endpoint.slice(API_VERSION.length).replace(/^\/+/, '');
 	}
+	return `${base}/${API_VERSION}/${endpoint}`;
```
oSumAtrIX: Defensive code like this isnt needed. Also the implementations of the server/worker side code differ from this.

src/lib/api/settings.ts
```diff
@@ -9,41 +10,115 @@ import {
...
+	if (value) {
+		localStorage.setItem(key, value);
+	} else {
+		localStorage.removeItem(key);
+	}
+}
+
+type StoredFallback = { url: string | null; recover: boolean };
+
+function readStoredFallback(): StoredFallback | null {
```
oSumAtrIX: It looks like the worker/server share similar code. Is it possible to have a shared code file where you can inject implementations of the persistence layer (e.g. the frontend would call the shared code with localstorage, server with cloudflare kv)

src/lib/api/settings.ts
```diff
@@ -9,41 +10,115 @@ import {
...
+}
+
+function syncLocal(key: string, value: string | null | undefined): void {
+	if (!browser) return;
+	if (value) {
+		localStorage.setItem(key, value);
+	} else {
+		localStorage.removeItem(key);
+	}
+}
```
oSumAtrIX: Weird name, simply call it save, or setStorage

.github/workflows/deploy.yml
oSumAtrIX: istg, wrangler.toml is so bad. Why do we have to do this, ive complained to cloudflare previously too. The namespace id is environment specific, so why do we have to set it in a file thats checked to source. 

Image

chatty says:

Image

This is good, do this instead with envsubst

mostafaNazari702: On the broader setup (template file + gitignored `wrangler.toml`) i can switch to that if you . i just went with the simpler in-place envsubst for now since it matched the "use envsubst instead" and we only need to inject a single env value anyway.

Do you want it split into a template setup?

src/lib/api/http.ts
```diff
@@ -0,0 +1,60 @@
...
+		const fresh = await fetchFn(url);
+		if (!fresh.ok) return fresh;
+
+		const cacheControl = fresh.headers.get('Cache-Control')?.toLowerCase() ?? '';
+		if (/\bno-store\b|\bno-cache\b|\bprivate\b/.test(cacheControl)) return fresh;
+
+		if (/\bmax-age\b/.test(cacheControl)) {
+			await this.cache.put(url, fresh.clone());
+		} else {
+			const body = await fresh.clone().arrayBuffer();
```
oSumAtrIX: Please use the cloudflare native cache as explained in my previous comments with overrriding the header if none

src/lib/api/server.ts
```diff
@@ -45,28 +72,64 @@ async function fetchJsonServer<T>(
...
+	fetchFn?: typeof fetch,
+	platform?: App.Platform
+): Promise<About> {
+	const storage = createStorageFromPlatform(platform);
+	const about = await fetchJsonServer('about', AboutSchema, fetchFn, storage);
+
+	if (storage) {
+		try {
+			if (about.fallback === null) {
+				await storage.delete(STORAGE_KEY);
```
oSumAtrIX: if you have .set, this should be .unset or simply do set(null)

src/lib/api/server.ts
```diff
@@ -45,28 +72,64 @@ async function fetchJsonServer<T>(
...
+	return fetchJsonServer('manager', ManagerReleaseSchema, fetchFn, storage);
 }
 
-export async function fetchAnnouncements(fetchFn?: typeof fetch): Promise<Announcement[]> {
-	return fetchJsonServer('announcements', AnnouncementsSchema, fetchFn);
+export async function fetchContributors(
+	fetchFn?: typeof fetch,
+	platform?: App.Platform
+): Promise<Contributable[]> {
+	const storage = createStorageFromPlatform(platform);
```
oSumAtrIX: Why do you create an instance of storage for every fetch instead of creating it once and passing it down?

